PR #15, opened 2026-09-09
fix(schedules): price committed B against the certified opening alphabet

RadNi:
## Summary

Custom schedules could price committed B for an honest base-4 decomposition (collision bound 3) while verification certified a base-8 witness alphabet (collision bound 7). Derive B's admission bound from the opening basis, consistently with frozen-group admission and planner sizing.

Add four public-API regressions covering rejection of the underpriced bound, a Q32 fixture that requires rank 2 instead of rank 1, acceptance and artifact round-tripping of correctly sized differing-base schedules, artifact decoding, and the bound exposed to security reports.

## Testing

- [x] `cargo test -p akita-config -p akita-schedules --lib --tests`
- [x] `cargo fmt --all --check`
- [x] `taplo fmt --check` and `typos`
- [x] All three exact Clippy configurations from `.github/workflows/ci.yml`, with `RUSTFLAGS='-D warnings'`
- [x] All five crate dependency-hygiene checks in the Clippy job
- [x] Rust file line-cap check and `git diff --cached --check`

## Security Considerations

- [x] Admission now rejects custom catalogs whose B coefficient bound is smaller than the certified opening alphabet requires. Correctly sized differing-base schedules remain admissible.
- [x] Rejection uses the existing `AkitaError::InvalidSetup` path; no verifier-reachable panic paths are added.
- [x] Transcript labels, challenge order, and domain separation are unchanged.
- [x] Artifact loading applies the corrected semantic audit without changing serialization or decode budgets.
- [x] No dependencies or CI actions are added.
- [x] Unsafe code is unchanged.

## Breaking Changes

Previously admitted custom catalogs that underprice B against the honest outer alphabet are now rejected. The 83 shipped rows use matching bases and require no artifact changes.

### DIFF
--- a/crates/akita-config/tests/committed_b_scanning_bound.rs
+++ b/crates/akita-config/tests/committed_b_scanning_bound.rs
@@ -1,0 +1,193 @@
+//! Public admission must price B against the certified opening alphabet.
+
+use akita_config::{
+    policy_of,
+    proof_optimized::{fp128, fp32},
+    CommitmentConfig, TrustedScheduleCatalog, ValidatedScheduleCatalog,
+};
+use akita_schedules::{planner_support::planned_next_witness_len, ResolvedScheduleRow};
+use akita_types::{
+    sis::num_digits_open, AkitaScheduleLookupKey, BlockGeometry, CommittedGroupBatchProfile,
+    DecompositionParams, FoldSchedule, GadgetDigits, OpenCommitMatrixParams,
+    OuterCommitMatrixParams, PolynomialGroupLayout, ScheduleSisBound, ScheduleSisRole,
+    TerminalResponseShape,
+};
+
+fn differing_basis_schedule<Cfg: CommitmentConfig>(
+    num_vars: usize,
+    b_bound: u128,
+) -> (CommittedGroupBatchProfile, FoldSchedule) {
+    let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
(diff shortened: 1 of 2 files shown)


quangvdao:
## Review

No actionable issues found at `d46bbd2f309b6a7b219c2c017135bc37e6c88e17` (merge base `fc4a034da801ab12d1ac6484a3e759f50fb90b4e`). Reviewed the full two-file diff.

The change correctly prices committed B against the opening alphabet actually enforced by verification. The shared audit covers root and recursive own-group commitments; frozen groups and setup prefixes already use the correct bound. B width appropriately remains based on the honest outer digit count. Correctly sized differing-base schedules remain admissible, and artifact loading repeats the corrected audit.

Validation:

- All 62 config/schedule tests passed locally with `cargo test -p akita-config -p akita-schedules --lib --tests`.
- Reverting only the production fix in an isolated checkout made three negative regressions fail because the old implementation accepted underpriced rows/artifacts. The positive corrected-schedule test still passed. Restored the head before the final test run.
- Local formatting, dependency hygiene, artifact identity, line-cap, Python and typo checks passed.
- Current GitHub checks passed; Book deployment was skipped. Full schedule regeneration and Clippy were checked in CI rather than rerun locally.

This addresses the committed-B admission finding from the audit. I consider this fix ready to merge within that scope; the separate estimator, constructor, cache and protocol audit items are unaffected.

mtbadakhshan (approved):

PR #20, opened 2026-09-09
feat(prover): simplify composable commitment execution

omibo:
This PR makes commitment execution composable across independently implemented backends. Applications define their backends and select the complete routing before execution; Akita handles source resolution, stage sequencing, state ownership, and result assembly.

The scope is **commitment only**. Integrating these backend interfaces into `prove()` is deferred.

**Independent commitment stages**

A commitment can use a different backend for each component:

```text
source → inner backend → outer backend → compression backend
```

Alternatively, a backend can implement inner and outer commitment as one fused operation:

```text
source → fused inner/outer backend → compression backend
```

The operation interfaces are:

- `InnerCommitOperation`
- `OuterCommitOperation`
- `FusedInnerOuterOperation`
- `CompressionOperation`

External crates can implement any of these operations and register them through `CommitmentExecutorBuilder`. A fused-only executor needs no dummy inner or outer implementation. Compression remains independently selectable in both routes.

Outer commitment and compression are **decoupled from the source representation**. They consume the preceding stage’s output and are registered on the executor, so they do not belong to `CommitmentSource`.

**Two ways to support a source**

A custom source can reuse Akita’s existing inner commitment implementations or provide its own algorithm.

For the standard path, the source implements `CommitmentSource`, advertises its available lossless representations through `available_polynomial_types`, and exposes the selected representation through `represent_as`:

| Representation | How a custom source maps to it |
|---|---|
| Dense | Expose canonical field coefficients, or cached balanced digit planes matching the selected inner plan. |
| One-hot | Expose the stored hot-position indices, chunk size, and variable count, preserving the one-hot structure without expanding it into dense coefficients. |
| Short-norm | Expose packed bounded signed coefficients with their encoding, lengths, and coefficient bounds. |

Akita selects a supported representation and runs the corresponding inner commitment implementation. A source may advertise multiple representations; only the selected one is borrowed or materialized.

For the external path, the source declares a backend-specific capability through `external_inner_commitment_capability` and supplies its operation through `prepare_external_inner_commitment`. This supports custom storage layouts and application-specific inner commitment algorithms without adding another built-in source representation to Akita. The operation must still satisfy Akita’s admission and commitment contracts.

A source can support both paths. Request compilation prefers a compatible homogeneous external capability for the source group; otherwise, it selects a supported standard path for the complete group. Unsupported combinations are rejected before arithmetic. External sources used with a fused backend must also support that backend’s fused request construction.

**Backend-defined retained state**

The retained hint no longer has to be a vector of ring elements. `BackendStateRef<K>` can own backend-specific state, including CPU witness buffers, a device allocation lease, or another owned resource handle.

Backends create state through `StateOwnerCapability::bind` and access it through the matching owner. Akita handles type erasure internally. Ordinary shared ownership and `Drop` replace the global state registry, token-to-buffer maps, generation counters, and composite dispatcher bookkeeping.

A fused GPU backend can therefore retain its large inner witness on the device throughout commitment. It returns an opaque state reference and the uncompressed commitment `u`. In this version, **`u` crosses the boundary as a host `RingVec<F>`** and is passed to compression; compression never receives the inner witness.

For split execution, inner and outer can share resident state directly when they share its owner. Crossing owners requires an explicit host-export connection. Portable state export is optional for resident fused commitment.

**Complete routing before execution**

Applications build the required executor combinations and declare the round ranges through `CommitmentExecutionScheduleBuilder`. Each round can combine different stage backends, and the selections can change from fold to fold. Inner/outer routing and compression have independent cutover points.

For example:

| Round | Inner/outer route | Compression |
|---|---|---|
| 0 | GPU fused | Metal |
| 1 | GPU fused | CPU |
| 2 | CPU inner + CPU outer | CPU |
| 3 | CPU inner + CPU outer | CPU |

The application registers every required combination, compiles the schedule, and can inspect every resolved step before execution. Compilation rejects gaps, overlaps, missing combinations, and fused/split mismatches without running a backend. Execution follows the selected route without fallback or reselection.

Per-round preflight additionally checks source admission, protocol mode, ring dimensions, and compression support. Rounds whose source metadata is available can be preflighted before the run starts.

Dynamic trait dispatch remains inside Akita, allowing runtime configuration while keeping the complete backend orchestration explicit in advance.

**Expected result and scope**

An external crate can implement its own inner, outer, fused, or compression operation, combine it with other external or built-in operations, and use the resulting executor for commitment. It does not need to implement Akita’s sequencing, type erasure, or composite state assembly.

This PR preserves the existing arithmetic plans, public commitment bytes, and transcript behavior. Coverage includes split/fused parity, explicit state-transfer boundaries, opaque-state ownership and cleanup, fused commitment without inner export, and independently scheduled backend changes.

New backend `prove()` methods, including device-resident opening and relation proving, are **not implemented in this PR**.


quangvdao:
I reviewed this at `b2b5a198`. The separation of inner, outer and compression operations, together with backend-owned retained state, looks useful. In particular, a fused commitment can retain the inner witness without exporting it to host. The spec also clearly separates that capability from future device-resident proving.

My main concern is how much configuration and orchestration surrounds those operations. I’d like to simplify that before extending the architecture further.

The design principle I’m proposing is to give each responsibility one owner:

- Sources describe the mathematical values and representations they can supply.
- Akita owns commitment geometry and protocol validation.
- A prepared backend implementation owns its operation, resources and supported representations.
- Execution connects those implementations and performs any required transfers.

This preserves the current mathematical stage boundaries without requiring every stage boundary to become a separate allocation, transfer or synchronization point. It also gives us a way to assess the added abstractions: each should own a distinct responsibility or eliminate duplicated work.

Here are the concrete places I’d revisit.

**1. Remove the obsolete commitment-backend generic.**

`ProverComputeStack` stores a `CommitmentExecutor`, but still carries `C: ComputeBackendSetup<F>` through `_commit_backend: PhantomData<fn() -> C>`.

Can we remove `C` from the stored stack type and keep any necessary generic on the constructor that builds the executor? The current arrangement propagates a backend type through callers and bounds even though the stack no longer stores that backend. [Code](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/stack.rs#L137).

**2. Consolidate route selection.**

`CommitmentExecutionScheduleBuilder` registers named pairs of complete executors. Meanwhile, `TieredProveStacks` remains the fold-level selector for production prover stacks. In this checkout, the new schedule builder’s Rust callers appear to be tests.

Independent inner/outer and compression cutovers are useful, but I’m not yet convinced they need another public registry of executor combinations. Could the cutovers resolve into the existing per-level selection? Alternatively, could this additional scheduler land with its first production consumer?

If a downstream consumer already needs it, including that usage would help establish what the existing selector cannot express. The concern is having two configuration layers responsible for overlapping decisions. [Schedule implementation](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/schedule.rs).

**3. Have preparation own correlated configuration.**

Registration supplies the operation, state owner, context, dimensions, capabilities and optional exporter through separate arguments. These describe one implementation, but the caller assembles them independently.

Could a prepared operation own this configuration and validate it once? The executor would then check compatibility between prepared operations.

This should replace existing registration plumbing rather than add another wrapper. The intended division is simple: an implementation establishes what it supports; the executor establishes whether the chosen implementations connect. [Builder](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/builder.rs).

**4. Allow consuming portable export to move CPU storage.**

`into_portable_hint(self)` calls the borrowed export path, and the CPU exporter clones the inner rows. Consequently, consuming resident state still copies the witness.

When the underlying CPU storage is uniquely owned, can conversion move those rows into the portable hint? Shared ownership may require a copy, but unique ownership should provide an opportunity to avoid one. This is a concrete case where improving the ownership boundary can also improve efficiency. [Conversion](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/state_policy.rs#L473), [CPU exporter](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/cpu/commitment_stage.rs#L38).

**5. Make the request-binding guarantee precise.**

`CommitmentStateBinding` contains setup, inner-plan geometry, source count and relation mode. Different requests with identical geometry therefore have equal bindings. However, `FullCommitmentOutput::new` describes its equality check as establishing that the outputs belong to the same request.

If the intended guarantee is detection of accidental cross-request mixing, the binding needs an invocation identity shared by that execution’s stages. Otherwise, the documentation and error should describe geometry/setup compatibility. This is a mismatch between the stated and implemented guarantee, not a demonstrated soundness exploit. [Binding](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/registration.rs#L19), [output check](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/stages.rs#L95).

I would prioritize these concrete changes over reducing lines to an arbitrary target. The useful outcome is fewer independently configured facts, fewer overlapping routing mechanisms, and fewer ownership-induced copies.

quangvdao:
A future consumer requirement worth documenting is several independent commitments derived from shared source data. I would leave its implementation outside this PR.

Aerie’s Falcon prover provides a concrete example. It makes eight separate commitments, several to interleaved views of common small-integer witness tables. The first five commitments precede a transcript challenge; that challenge is used to produce the three JL commitments. Its current commitment adapter converts the compact views into dense field polynomials.

There are two potential improvements: preserving compact representations and sharing production across independent commitment requests. Neither requires packing those requests into one polynomial.

The important distinction is:

> Source ownership, computation sharing and commitment identity are independent.

One producer can supply several polynomial views. Each view still has its own mathematical meaning, and each commitment request retains its own profile, ordering and opening obligations.

The current `CommitmentSource` interface selects a representation or external inner operation within a request. It does not directly expose the relationship between several requests that share production. [Source interface](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/source.rs#L483).

A possible future design would have three extension points:

- **Source description:** expose logical polynomial outputs, their available representations, and any shared producer or storage.
- **Backend preparation:** select an implementation for the requested computation, with visibility into related requests before expensive materialization.
- **Executable dispatch:** execute the prepared work and return owned results.

These responsibilities could eventually become three traits. I would not introduce unused traits or a general computation graph in this PR. A short design sketch showing how the current interfaces could evolve is sufficient.

For Aerie, the intended process would be straightforward: describe the outputs available in the current phase, request their separate commitments, prepare and execute them together where useful, then collect the commitments in protocol order. After deriving the challenge, repeat for the JL outputs.

The extension should preserve several properties:

- A single-output source remains easy to implement.
- Multiple requests can reference shared ownership without copying the underlying data.
- Representation selection happens before expensive materialization.
- Custom source-dependent kernels remain possible.
- Sharing does not assume identical profiles, a common matrix, or that one traversal is always faster.
- Challenge-dependent work stays in its proper protocol phase.

In particular, the current source-owned external operation can remain a working integration mechanism. The design note should explain how a future preparation step could coordinate several such outputs, rather than permanently requiring each source to materialize or submit its work independently.

I would keep this PR focused on simplifying the existing commitment implementation. The goal of documenting this example is to avoid closing off a useful extension, not to make shared-source batching, distributed execution or device-resident proving prerequisites for this change.

quangvdao:
One remaining limitation: outer/fused commitment returns `u` as a host `RingVec`, so it must reach host memory before compression, even if the selected compressor could consume it on device.

CPU compression may be preferable at these sizes; I’m not claiming this transfer is a measured bottleneck. But ideally that would be an execution choice rather than an interface requirement.

Could the existing backend-owned value mechanism cover `u` too, without adding much machinery? If that isn’t a small, clean change, I’d document the limitation and defer it.

quangvdao:
# Akita PR #20 follow-up review

Reviewed head: `b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f`.
Base/merge base: `029ceceaba879a0d9bf80a47d2e08534e630457d`.

This follow-up review covers protocol integration, backend boundaries, and tests/specification/compatibility. It consolidates three agent review passes with a separate cross-check of the code evidence. It excludes the concerns already raised in my earlier comments, except where a distinct failure is established. No production code was changed.

## PR Review findings

### 1. Moderate — malformed custom one-hot positions pass representation validation

A source declaring chunks of size 8 can supply `Some(9)`. The materialization boundary checks the representation's shape and number of positions but not that each position is inside its chunk. The built-in constructor checks this invariant; the new custom-source path bypasses that protection. CPU traversal computes `chunk_index * chunk_size + index`, so an invalid index can spill into another chunk or be omitted outside the processed range.

A focused rejection test failed at this head because materialization returned `Ok`. In an isolated checkout, the test used a custom source with `K = 8` and first position `Some(9)`, then asserted that `compile_commitment_request(...).materialize()` rejects it. The scoped command `rtk cargo test -p akita-prover --lib malformed_custom_onehot_position_is_rejected_at_materialization --no-default-features --features transcript-blake2b -- --nocapture` exited 101: one failed test, 419 filtered. The test was a scratch reproduction added for this review, not an existing upstream test. This is a malformed-source validation defect; no invalid-proof acceptance was established. Restore the invariant at a checked representation boundary or validated traversal without introducing an unnecessary copy or repeated full scan.

Evidence: [source.rs:640](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/source.rs#L640), [onehot/poly.rs:65](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/backend/onehot/poly.rs#L65).

### 2. Moderate — returned state is not compared with the executor's request binding

The executor creates a binding containing its setup and actual source count, but does not compare backend results against it before forwarding them. Later equality checks compare backend-produced states to each other, and compression receives the returned image's binding. A result carrying a different valid setup or source count can therefore remain self-consistent and pass through.

This is stronger and different from the posted observation that bindings lack invocation identity: even the fields already present are not enforced against the request. Compare each returned binding with the executor's expected binding before invoking the next stage. This is backend-result validation, not evidence of a malicious backend defeating the verifier.

The unchecked return occurs in [executor.rs:494–514](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/executor.rs#L494-L514). The split path passes the image to outer execution, and the fused path returns directly ([executor.rs:553–596](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/executor.rs#L553-L596)). The output constructor checks the inner plan and `u` dimensions, but not the expected setup/source count ([stages.rs:46–61](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/stages.rs#L46-L61)). Full execution then supplies `image.binding()` to compression ([executor.rs:636–644](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/executor.rs#L636-L644)).

**Static counterexample:** a custom operation returns state with the requested inner plan and relation mode, but another valid setup descriptor or nonzero source count. Inner-only state assembly accepts it; uncompressed output validation accepts it; full execution can propagate the same incorrect binding into compression. This contradicts the [stage-boundary validation contract](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/specs/composable-commitment-execution.md#L97-L99).

**Regression to add:** backends returning wrong setup/count bindings, with counters proving rejection before outer execution, export or compression. Pass the executor's expected binding to compression and validate its result against that authority.

### 3. Moderate — mixed fused/split configurations prepare and report the wrong caches

Direct full/uncompressed execution chooses the fused implementation when registered. Proof-wide preparation and accounting choose fused resources only when split registrations are absent. A configuration supporting fused nonterminal commitment and split terminal commitment therefore prepares/reports split A/B resources for work that will actually use fused resources.

Unused split caches can coexist with subsequently allocated fused caches, and memory estimates identify the wrong owners. Resolve the route once for both preparation and execution; merely preferring fused everywhere would break terminal inner-only routing.

Compare direct selection in [executor.rs:212–238](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/executor.rs#L212-L238) with proof-wide selection in [resources.rs:394–422](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/resources.rs#L394-L422). Prewarm, retained-owner selection and planned bytes all use the latter ([resources.rs:425–461](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/resources.rs#L425-L461)). Production proving calls proof-wide prewarm at [prove.rs:98](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/protocol/core/prove.rs#L98).

**Static counterexample:** register distinct split-inner, split-outer and fused resource owners. For nonterminal A/B requirements, proof-wide prewarm targets the split owners; executing the corresponding request subsequently prepares the fused owner's resources. `RoutedNttRequirement` carries a stage but not the execution mode or resolved route, so the current helper cannot distinguish this from terminal split-inner work.

**Regression to add:** one mixed executor spanning nonterminal fused A/B and terminal split inner; assert the precise resource owners prepared and reported for each. Use the same canonical routing decision for preparation, accounting and execution.

### 4. Moderate — downstream terminal-state implementations cannot construct the result carrier independently

The public `TerminalBindingState` trait returns `TerminalTFieldsMessage`, but that type has no public constructor and its only field is private. Downstream implementations must delegate through a built-in state/hint to obtain it. A downstream-style compile probe reproduced E0451.

Expose a checked semantic constructor while keeping canonical transcript encoding library-owned. This is a public extension API defect, not a demand for GPU proving in this PR.

The public proving API requires this trait at [prove.rs:65–67](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/protocol/core/prove.rs#L65-L67). The return type and trait are at [state_policy.rs:303–318](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/state_policy.rs#L303-L318); construction helpers are private ([state_policy.rs:321–365](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/state_policy.rs#L321-L365)).

**Reproduction:** an isolated downstream-style integration test attempted `Ok(TerminalTFieldsMessage { bytes: Vec::new() })`; `rtk cargo check -p akita-prover --test external_terminal_state_api` failed with `E0451: field bytes of struct TerminalTFieldsMessage is private`. The empty vector was a compile-time access probe, not proposed valid transcript data.

**Regression to add:** an external state implementation that creates the result from a valid terminal row through a checked public API, without constructing a built-in hint. Canonical encoding should remain Akita-owned.

### 5. Moderate — unsupported state consumption is detected after transcript mutation

The active spec promises proof-route capability preflight before transcript mutation. `batched_prove` instead binds the instance descriptor before discovering that a resident state lacks the exporter/consumer required by the current proving path. Commitment with that resident state is allowed to succeed, so this is a reachable unsupported-route failure.

Check the required state-consumer capabilities before changing the caller's transcript. Do not materialize witness rows merely to check support. The absent GPU proving implementation remains deferred; the finding is about when an unsupported route is rejected.

The requirement is explicit in [the spec](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/specs/composable-commitment-execution.md#L390-L393). The descriptor is bound at [prove.rs:90–106](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/protocol/core/prove.rs#L90-L106), while inner state is consumed later at [ring_relation.rs:439–445](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/protocol/ring_relation.rs#L439-L445).

**Static failure path:** commit through an allowed fused resident route without an inner exporter, then supply its state to the current proving adapter. `inner_relation_material()` returns `commitment route has no inner-relation state operation` ([state_policy.rs:377–388](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-prover/src/compute/commitment/state_policy.rs#L377-L388)), after the descriptor has already been absorbed into the caller's transcript.

**Regression to add:** a logging transcript and a state whose required consumer is unsupported; assert capability rejection before any transcript event. Include the scheduled compression and terminal capabilities in the proof-route assessment. This is a preflight check, not a request to implement those unsupported backends.

### 6. Moderate — current getting-started and backend documentation uses the removed API

The compiled example was migrated, but the Book quickstart and commitment API chapter still pass `&stack` to `commit` and access `commit_output.hint`. The API now takes `stack.commitment()` and returns `prover_state`. The PR-modified compute-backend guide also retains the old call. Rendering these snippets does not check compilation.

Update the current user journeys alongside the API, preferably sourcing snippets from the compiled example. This is stale documentation at the reviewed head, not a request for backwards compatibility.

The new signature is at [scheme/mod.rs:132–151](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/crates/akita-pcs/src/scheme/mod.rs#L132-L151). Stale instructions remain in [docs/compute-backends.md:27–46](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/docs/compute-backends.md#L27-L46), [the Book quickstart:75–117](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/book/src/usage/quickstart.md#L75-L117), and [the commitment API chapter:8–37](https://github.com/LayerZero-Labs/akita/blob/b2b5a198ef4fcb90d88c96c2f9ddec1cbbff760f/book/src/usage/commitment-api.md#L8-L37). The compiled `crates/akita-pcs/examples/quickstart.rs` already demonstrates the migrated API.

## Specification alignment

- Opaque owned state, fused commitment without inner export, compact source discovery, and existing CPU proving adapters are implemented.
- Built-in resident-state end-to-end tests cover compressed/raw recursive levels and terminal verification; those paths are not simply untested.
- Backend-result boundary validation, complete proof-route preflight, and preparation/execution resource agreement have the gaps above.
- Device-resident proving, shared-source batching and distributed implementations remain deferred. They are not additional findings.

The ownership and backend-operation boundaries are earned abstractions. The useful new simplification target is to give the resolved execution route one owner and make request metadata authoritative at the executor boundary. Adding wrappers, a host/device enum specifically for u, or a generic graph engine is not necessary to fix these findings.

## Validation and limits

- One malformed-source rejection test ran and failed as expected because the current code accepts it.
- A downstream terminal-message compile probe failed with the expected privacy error.
- Remaining findings use static call-path evidence, not completed runtime reproductions.
- No full local workspace test suite or new performance measurements were run.
- Final observed PR head unchanged. GitHub rollup: 47 successful checks and one skipped book deployment.
- No verifier-acceptance defect was established. Review findings should not be described as cryptographic exploits.

quangvdao:
Follow-up review of **759c60682fbf760afc4a2ac2f3965d0aa6551539**, against base **38dbf031013a2b8f7ff93252e190dac2b3353706**. This updates the [previous review](https://github.com/LayerZero-Labs/akita/pull/20#issuecomment-5611025573).

This revision is materially better. Most previously reported issues are fixed, and the prepared-stage and homogeneous-group changes are useful simplifications. The remaining findings below concern setup compatibility, terminal routing, retained-state validation, and one source-shape edge case. Each includes a suggested correction; the earlier findings are accounted for separately below.

## Remaining findings

### 1. Major — components prepared for one setup can be accepted under another

Prepared stages can be created under builder A and then registered with builder B. `operation_context` checks resources against A, but the receiving registration/build path never checks them against B. The built-in CPU operations continue computing with A while the executor constructs bindings using B.

A scratch public-API reproduction registered CPU stages prepared under A in builder B. `build` and `execute_full` succeeded, but the returned commitment differed from a correctly configured B executor. This is an actual wrong-result case for a misconfigured supported caller, not a verifier exploit.

The same missing setup-identity comparison exists where root `commit` and setup-prefix commitment receive both an expanded setup and an independent executor, and where `ProverComputeStack::new` accepts the commitment executor while validating only the other three operation contexts.

Correction: give the executor one checked setup-compatibility operation and use it at these composition boundaries. Validate each prepared stage's resource setup when adopting it into a builder. Existing source/output-binding checks must remain, but they cannot establish that the configured arithmetic uses the caller's setup.

Evidence: [builder](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/commitment/builder.rs#L46-L156), [stack assembly](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/compute/stack.rs#L172-L190).

### 2. Moderate — the standalone schedule cannot express valid inner-only and fused-plus-inner executors

The executor builder supports an inner-only executor, and a fused-plus-inner executor for fused nonterminal commitment followed by terminal inner-only commitment. However, schedule validation allows `Split` only when the executor has an outer operation and no fused operation. Labeling a mixed executor `Fused` instead makes terminal preflight/execution reject it.

A full split CPU executor can execute terminal work through the schedule. The defect is that callers must provide an unused outer operation and/or a separate terminal executor to represent configurations the builder explicitly supports.

Correction: validate the actual operation required by the round, or represent inner-only routing explicitly. Do not simply relax `Split` to accept mixed executors: full execution would still prefer fused and contradict the named route. Add a regression covering fused nonterminal plus split terminal execution with no dummy outer.

Evidence: [kind matching](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/commitment/executor.rs#L116-L123), [terminal schedule behavior](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/commitment/schedule.rs#L102-L157).

### 3. Moderate — proof preflight still does not check the requested retained-state plan

Missing exporters now fail before transcript mutation. However, preflight takes no expected plan and resident implementations check exporter availability only. A state retained from a different valid commitment can pass these checks, then fail when runtime materialization checks the actual compression plan/relation mode, after transcript absorption.

This is the remaining part of the previous late-failure finding, not an unchanged repetition of it. Check expected per-group metadata before transcript work, without exporting the witness just to inspect support. This is a metadata compatibility check; it need not establish correctness of arbitrary backend arithmetic.

Evidence: [preflight calls](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/protocol/core/prove.rs#L90-L128), [resident preflight and later checks](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/commitment/state_policy.rs#L529-L576).

### 4. Moderate — oversized one-hot chunks can still produce an accepted empty representation

The previous out-of-range-index example is fixed. A separate shape edge remains: a custom source declaring 64 coefficients, chunk size 128 and an empty position slice passes compilation/materialization because the expected count is computed as floor division, `64 / 128 = 0`. This does not represent the exact complete-chunk layout enforced by the built-in constructor.

A focused scratch negative test reproduced acceptance at 65cb41f; this validation code is unchanged at 759c6068. Require the logical coefficient count to be an exact multiple of the nonzero chunk size and check the exact position count. If partial chunks are desired instead, they need an explicit mathematical/encoding contract; silently flooring the count is not that contract.

Evidence: [source validation](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/commitment/source.rs#L661-L690), [built-in constructor](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/crates/akita-prover/src/backend/onehot/poly.rs#L184-L207). No invalid-proof acceptance was established.

## Documentation suggestions

The architecture table still describes an application-defined enum polynomial carrier for heterogeneous representations, while this head removes the built-in enum and introduces whole-group erasure. An application-defined enum remains supported by the generic bounds, so this is a guidance cleanup rather than an API defect: distinguish homogeneous `PreparedProverGroup` from heterogeneous batches assembled through `ErasedPreparedProverGroup`, and describe the intended path first.

Also reconcile the active spec's references to planned byte accounting with the deliberately removed unused NTT planning APIs, and clarify the PR body: commitment executors are already used by the existing proving path; new proving-operation implementations are what remain deferred.

Evidence: [architecture table](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/book/src/how/architecture.md#L141-L142), [correct new explanation](https://github.com/LayerZero-Labs/akita/blob/759c60682fbf760afc4a2ac2f3965d0aa6551539/book/src/usage/commitment-api.md#L176-L192).

## Late merge from main

The full 11-file merge delta to 759c6068 was also reviewed. It moves claim point/evaluation checks into opening-data construction, centralizes schedule admission, and removes public low-level proving entry points that bypass that admission. These are useful simplifications. No additional actionable protocol finding was identified in the merge. It does not change the commitment setup/schedule/source paths, and the retained-state preflight limitation remains.

## Previous review and suggestion status

| Earlier concern | Latest assessment |
|---|---|
| Out-of-range one-hot indices | Fixed for the reported example and all stored index widths. Separate oversized-chunk case remains above. |
| Backend results ignore expected binding | Fixed with invocation identity and owner/binding checks. Composition across different setups remains a distinct problem. |
| Wrong fused/split cache routing | Fixed in retained/prewarm routing; unused planned-byte APIs removed. |
| Terminal result cannot be constructed externally | Fixed with public `from_row`; external integration test passes. |
| Late unsupported state failure | Fixed for absent exporters; incomplete for concrete retained-state/plan compatibility. |
| Old quickstart/API examples | Fixed at reported locations; smaller architecture/spec cleanup remains. |
| Phantom commitment backend generic | Removed. |
| Correlated stage configuration | Bundled into prepared stages, but receiving setup validation remains necessary. |
| Consuming export copies unique state | Fixed for unique storage; shared leases use the copying fallback. |
| u and shared-source batching | Deferred; reasonable to keep outside this PR. |

## Suggested architectural direction

The changes are substantive improvements. Keeping concrete types within each group and erasing whole groups preserves heterogeneous opening batches without a built-in per-polynomial sum type. Removing unused planning APIs is preferable to maintaining a second model with no consumer. The new prepared-stage boundary is useful, provided setup compatibility is actually checked when connecting stages.

I would fix the concrete composition seams above before introducing more abstractions. No generic graph engine, device/host enum for u, or new batch API is required here. A single authority for setup compatibility and mode-aware route validation should simplify both runtime behavior and validation.

## Performance and validation

Eight focused checks passed on the final pristine merge head (the original five also passed before the merge):

- executor rejects changed invocation/foreign state owners;
- consuming extraction moves unique state and preserves shared state;
- one-hot bound validation covers all index widths;
- resolved homogeneous groups cover dense and one-hot widths;
- external crate composes fused and compression backends;
- opening-data construction rejects group arity mismatch;
- opening-data construction rejects misaligned claim material;
- transcript binding rejects malformed commitment rows.

Two separate scratch failing regressions were run at 65cb41f demonstrating setup mismatch and malformed oversized chunks. Their commitment implementation paths are unchanged at 759c6068. Protocol preflight and standalone schedule findings are static call-path proofs, not completed runtime reproductions. No full local suite or fresh benchmark was run.

Final observed head: 759c6068. GitHub: 47 successful checks, one skipped Book deployment.

The benchmark inspected for pre-merge head 65cb41f reports Fp64 one-hot commitment +6.4%, while Fp128 one-hot direct/recursive commitments are -8.4%/-10.0%. Only three samples were measured after warm-up. Before accepting performance claims, rerun the Fp64 outlier and the W4R2/W8R2 cases above 3% with at least the spec's ten samples and unchanged cache/input conditions. These numbers do not establish a uniform speedup, slowdown, or cause. [Benchmark report](https://github.com/LayerZero-Labs/akita/pull/20#issuecomment-5610374006).

crates/akita-prover/src/commitment/external.rs
```diff
@@ -0,0 +1,388 @@
...
+pub fn cpu_external_inner_prepared_setup<F: Field + 'static>(
+    context: &dyn Any,
+) -> Result<&CpuPreparedSetup<F>, AkitaError> {
+    context
+        .downcast_ref()
+        .ok_or_else(|| AkitaError::InvalidInput("external CPU commitment context mismatch".into()))
+}
+
+/// Open, process-local backend family identity.
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
```
RadNi: **Correctness:** `BackendKindId` derives `Eq`/`Hash` over both the marker `TypeId` and the diagnostic `name`. That makes two values for the same backend marker compare unequal solely because their display labels differ, which can reject otherwise homogeneous external-source groups. Please make equality/hash depend only on the marker identity (manual impls are enough) and keep `name` for diagnostics. The same issue applies to the derived equality of `ExternalInnerCommitmentCapability`, whose `diagnostic_name` should not change compatibility. A regression test with identical marker types and different labels would pin this down.

crates/akita-prover/src/commitment/source.rs
```diff
@@ -0,0 +1,1499 @@
...
+    /// Construct a descriptor after checking its extents.
+    pub fn new(
+        num_vars: usize,
+        total_coefficient_len: usize,
+        live_coefficient_len: usize,
+        class: CommitSourceClass,
+        family_name: &'static str,
+    ) -> Result<Self, AkitaError> {
+        let logical_len = checked_logical_len(num_vars)?;
+        if live_coefficient_len == 0
```
RadNi, line 49: **Contract mismatch / late failure:** this admits `live_coefficient_len < logical_len` for every source class, and `validate_plan_extents` prices the request from that live extent. The dense CPU path later recomputes its ring count from `2^num_vars`, however, so a custom dense source can pass source compilation and then fail only inside backend execution because its logical block count differs from `plan.num_live_blocks`. Please encode the class-specific rule at this boundary (for example, require dense/one-hot live extent to equal logical extent), or make materialization and the kernels consistently honor the shorter live prefix. That gives external sources one clear, early-validated contract.

crates/akita-prover/src/commitment/executor.rs
```diff
@@ -0,0 +1,1431 @@
...
+                return Err(AkitaError::InvalidInput(
+                    "compression operation does not support the selected dimensions or relation mode"
+                        .into(),
+                ));
+            }
+        }
+        Ok(())
+    }
+
+    fn validate_setup_capacity(&self, plan: &CommitmentExecutionPlan) -> Result<(), AkitaError> {
```
RadNi, line 335: **Duplicate security/sizing logic:** this independently reconstructs the A, B, and compression setup footprint even though setup-envelope construction already computes the commitment footprint. The formulas agree now, but any future change can make allocation and commit-time admission disagree. Please expose one checked `CommitmentExecutionPlan::max_setup_field_elements()`-style primitive and use it from both setup construction and this validation. This removes arithmetic from the executor and prevents split-brain sizing without changing the hot path.

crates/akita-prover/src/commitment/executor.rs
```diff
@@ -0,0 +1,1431 @@
...
+struct CpuInnerContext;
+
+/// Checked commitment-stage executor.
+///
+/// Registrations keep each operation with its capabilities and explicit
+/// export edge. Stage outputs directly own their backend values, and the
+/// selected state policy assembles those values without a state registry.
+pub struct CommitmentExecutor<
+    'a,
+    F: Field + CanonicalEncoding,
```
RadNi, line 48: **Simplification:** four independent `Option<Prepared…>` fields encode several impossible configurations, so the builder and nearly every executor method have to rediscover whether the route is inner-only, split, fused-only, or split-with-fused. Compile the builder into one route enum, e.g. `InnerOnly { inner }`, `Split { inner, outer, fused: Option<_> }`, and `FusedOnly { fused }`, with compression kept orthogonal. Full execution can still prefer `fused`, inner-only still uses `inner`, and trait-object dispatch/kernel performance is unchanged. This would delete much of `split_inner`, `split_outer`, `fused_for_plan`, `matches_inner_outer_kind`, and the repeated invalid-state checks while preserving all supported configurations.

crates/akita-prover/src/commitment/schedule.rs
```diff
@@ -0,0 +1,608 @@
...
+        if name.is_empty() || range.start > range.end || range.end > round_count {
+            return Err(AkitaError::InvalidInput(
+                "commitment route has an invalid name or round range".into(),
+            ));
+        }
+        Ok(())
+    }
+
+    /// Resolve every independently assigned pair into an immutable plan.
+    pub fn compile(self) -> Result<CommitmentExecutionSchedule<'a, F, SP>, AkitaError> {
```
RadNi, line 314: **This is a second routing system on top of `CommitmentExecutor`:** each round is matched through string names and an A/B-compression pair, and every round must name a compressor even for `InnerOnly`/`Uncompressed` plans where none is executed. Within this PR the public schedule has no non-test production consumer, while the proving stack already owns the per-round executors. The simplest implementation is to defer/remove this registry and store the chosen executor directly per round. If named schedules must remain, model the step as mode-specific enum variants (`InnerOnly`, `Uncompressed`, `Full { compression }`) and validate concrete operation availability at compile time. Either approach removes dummy compressor assignments and preserves route choice with no effect on commitment kernels.

crates/akita-prover/src/commitment/source.rs
```diff
@@ -0,0 +1,1499 @@
...
+        }
+    }
+    Ok(())
+}
+
+/// Compile source paths for one checked inner plan.
+///
+/// Discovery and capability intersection complete for the entire source group
+/// before any standard or external representation is materialized.
+pub fn compile_commitment_request<'a, F: Field>(
```
RadNi: **Maintainability:** this new module is 1,499 lines and combines the public source/representation model, selection compiler, extent validation, built-in source adapters, and roughly 430 lines of tests. That concentration makes the extension boundary much harder to audit than necessary. Please split it into focused modules such as `descriptor`, `representation`, `compiler`, `builtins`, and `tests`; keep `compile_commitment_request` as the single orchestration entry point. This is a source-only reorganization, so it preserves every feature and has no runtime cost.

crates/akita-prover/src/compute/cpu/commitment_stage.rs
```diff
@@ -0,0 +1,569 @@
...
+                        Ok(witness.inner_rows)
+                    })
+                    .collect()
+            }
+            Err(shared) => self.export_rows(plan, &shared),
+        }
+    }
+}
+
+struct CpuInnerImageExporter<F: Field> {
```
RadNi, line 237: **Redundant wrapper:** `CpuInnerCommitOperation` already implements `InnerImageExportOperation` with these same two forwarding calls, so `CpuInnerImageExporter` is a second object and duplicate implementation around the same store. Register `inner_operation.clone()` directly as `Arc<dyn InnerImageExportOperation<F>>` and delete this type plus `portable_exporter()`. This preserves the separate export capability and ownership behavior while removing an allocation and one layer of indirection.

crates/akita-prover/src/compute/cpu/compression_stage.rs
```diff
@@ -0,0 +1,306 @@
...
+                        )
+                    })?,
+            )
+            .ok_or_else(|| {
+                AkitaError::InvalidSetup("CPU compression retained bytes overflow".into())
+            })
+    }
+}
+
+struct CpuCompressionExporter<F: Field> {
```
RadNi, line 54: **Same exporter duplication as the inner stage:** move these two methods onto `CpuCompressionOperation` (which already owns the capability) and register a clone of the operation as `Arc<dyn PortableCompressionStateExport<F>>`. The exporter object currently carries only a cloned owner and adds no independent behavior. Removing it keeps portable/consuming export semantics unchanged and reduces the number of backend-facing types.

crates/akita-prover/src/commitment/state_policy.rs
```diff
@@ -0,0 +1,794 @@
...
+    }
+
+    /// Consume into the public payload and prover-private state.
+    pub fn into_parts(self) -> (RingVec<F>, S) {
+        (self.terminal_payload, self.prover_state)
+    }
+}
+
+/// Policy-selected state output for an inner-only commitment route.
+pub struct CommitmentStateOutput<S> {
```
RadNi: **Pass-through type:** `CommitmentStateOutput<S>` contains only `S`, and the production caller immediately consumes it with `into_state()`. Returning `SP::State` directly from inner-only execution would remove this public struct and its accessor/constructor plumbing without losing metadata or validation—the policy binding has already completed before this wrapper is built. That also makes inner-only execution consistent with what callers actually need.

crates/akita-prover/src/commitment/external.rs
omibo: Fixed in 08f1a5e09. BackendKindId equality and hashing now use only the backend marker TypeId, and ExternalInnerCommitmentCapability equality ignores diagnostic_name. I also added a regression test covering identical marker types with different labels, including HashSet deduplication.

crates/akita-prover/src/commitment/source.rs
omibo: Fixed in 08f1a5e09. CommitSourceDescriptor now requires dense and one-hot live extents to equal the full logical extent. ShortNorm remains the only class allowed to expose a shorter live prefix. A regression test covers all three classes.

crates/akita-prover/src/commitment/executor.rs
omibo: Fixed in 08f1a5e09. The checked A/B/compression maximum now lives in akita-types as commitment_execution_setup_field_elements. Setup-envelope construction and CommitmentExecutionPlan::max_setup_field_elements both call that primitive, and the executor only compares the resulting value with setup capacity.

crates/akita-prover/src/commitment/executor.rs
omibo: Fixed in 08f1a5e09. CommitmentExecutor now stores one PreparedInnerOuterRoute enum plus orthogonal optional compression. The variants are InnerOnly, Split, and Fused with an optional terminal inner operation. The builder rejects simultaneous fused and split-outer registration, while preserving fused nonterminal rounds followed by terminal inner-only execution.

crates/akita-prover/src/commitment/schedule.rs
omibo: Addressed the concrete mode mismatch in 08f1a5e09. Compression selection is now optional, and None is an explicit scheduled no-compression decision distinct from an unassigned round. InnerOnly and Uncompressed plans require None; Full plans require a compressor. I retained the named commitment-only schedule because this API must support independently predeclared A/B and compression cutovers. The proving path continues to use TieredProveStacks as its routing authority.

crates/akita-prover/src/commitment/source.rs
omibo: Fixed in 08f1a5e09. The source code is split into the public source and representation contract, compiler.rs for selection and validation, builtins.rs for Akita adapters, and tests.rs. compile_commitment_request remains the single orchestration entry point. The main source.rs file is now 559 lines.

crates/akita-prover/src/compute/cpu/commitment_stage.rs
omibo: I verified this suggestion but did not apply it. CpuInnerCommitOperation borrows CpuBackend and CpuPreparedSetup, while the exporter is retained inside policy-selected state after execution and its trait object is effectively static. Coercing the operation Arc directly would either fail the lifetime requirement or force the retained state API to carry the executor lifetime. CpuInnerImageExporter deliberately owns only the cloned StateOwnerCapability, so exported state can outlive the borrowed operation safely.

crates/akita-prover/src/compute/cpu/compression_stage.rs
omibo: I verified this case and kept the separate exporter for the same lifetime reason. CpuCompressionOperation borrows backend, prepared setup, and expanded setup, while portable compression export can be retained in policy state beyond that operation borrow. CpuCompressionExporter owns only the state-owner capability and is therefore independent of those borrowed resources.

crates/akita-prover/src/commitment/state_policy.rs
omibo: Fixed in 08f1a5e09. CommitmentStateOutput was removed, execute_inner and the schedule now return SP::State directly, and the terminal ring-switch caller no longer immediately unwraps a pass-through value.

RadNi (approved):

quangvdao:
## PR Review findings

Reviewed exact surface: `c288f487fff4dba8269731da613284b594aeb545...05b17d191260f1ed837245cbc7273db21114f738`.

### 1. Major — caller-supplied short-norm extrema can select an invalid commitment kernel

[`ShortNormRepresentation::new`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/commitment/source.rs#L295-L341) forwards caller-supplied `negative_abs_max` and `positive_max` into `PackedSignedDigitView`. [`from_encoded`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/backend/packed_digits/mod.rs#L403-L440) checks that those claimed extrema fit the stored bit width, but never compares them with the decoded coefficients. The selected-representation compiler also checks only extents and non-emptiness ([`compiler.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/commitment/source/compiler.rs#L109-L117)).

**Concrete failure:** construct the public representation with encoded bytes `[0x03]`, live length `1`, physical length `64`, signed width `3`, and claimed extrema `(0, 1)`. Construction succeeds, although the first decoded coefficient is `+3`. Under a one-digit plan with `log_basis_inner = 2`, the CPU path trusts the false bounds and selects the balanced kernel ([`commitment.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/compute/cpu/commitment.rs#L420-L465)). Debug builds hit the balanced-LUT range assertion. On LUT-backed release paths, masking maps `+3` to the entry for `-1`: `(3 + 2) & (4 - 1) = 1` ([`lut.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-algebra/src/ring/crt_ntt_repr/lut.rs#L150-L165)), producing a commitment to different coefficients.

**Required correction:** derive the exact extrema by scanning the live encoded coefficients during selected-representation materialization, or reject when derived and supplied extrema differ. Only verified extrema should control balanced/raw dispatch. Add a regression using the payload above that rejects before arithmetic in debug and release configurations.

### 2. Moderate — terminal `InnerOnly` routing is lost before cache-policy evaluation

`NttExecutionRequirements::add_terminal` correctly emits an `InnerOnly` request ([`requirements.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/compute/requirements.rs#L477-L503)). `CommitmentExecutor::retained_routed_requirement` uses that route to select split-inner resources, but then converts it to `CommitmentNttRequirement`, which has no route field. [`CommitmentNttRequirement::routed`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/commitment/resources.rs#L66-L116) reconstructs every request as `InnerOuter` before calling the backend's cache policy ([`resources.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/commitment/resources.rs#L230-L250)).

A backend permitted to stream terminal A while caching nonterminal A/B will therefore classify the terminal operation as cached and prewarm/build a slot that runtime policy intended to stream. This contradicts the active spec's requirement that one resolved route control prewarm, owner selection, and execution.

**Required correction:** carry `CommitmentNttRoute` in `CommitmentNttRequirement`, or pass the original `RoutedNttRequirement` unchanged through `CommitmentResourceControl`. Add a route-sensitive backend test asserting that terminal retention and prewarm both observe `InnerOnly`.

### 3. Moderate — resident inner exports are not fully validated or frozen at their boundary

[`ResidentCommitmentState::inner_relation_material`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/commitment/state_policy.rs#L512-L545) invokes the backend exporter and then constructs `InnerRelationStateMaterial`. That constructor checks only that each row is divisible by the ring dimension ([`state_policy.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/commitment/state_policy.rs#L341-L359)); it does not enforce `rows.len() == binding.source_count()` or the exact per-row length `num_live_blocks * n_a * ring_dimension`.

For example, with `D=64`, `num_live_blocks=1`, `n_a=2`, and one source, an exporter can return one 64-coefficient row. It passes the boundary constructor although 128 coefficients are required. The exact length is rejected only later during ring-switch construction ([`coeffs.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/protocol/ring_switch/coeffs.rs#L397-L419)), after transcript-dependent work.

There is also a latent consistency failure for stateful custom exporters: the terminal suffix exports once to construct and absorb `terminal_t_fields_message`, then exports again to obtain the proof row ([`suffix.rs`](https://github.com/LayerZero-Labs/akita/blob/05b17d191260f1ed837245cbc7273db21114f738/crates/akita-prover/src/protocol/core/suffix.rs#L270-L276)). Built-in CPU exporters are pure, but the public trait does not enforce purity; alternating same-shaped rows would bind row A and prove with row B.

**Required correction:** introduce one canonical constructor/helper that takes the binding and exported rows, checks ring dimension, source count, and exact checked coefficient lengths, and returns frozen material. Terminal binding and relation construction should derive from the same validated export.

## Specification alignment

- The independent inner/outer/compression stages and backend-owned retained state are implemented as real boundaries.
- Source admission is incomplete for caller-supplied packed extrema.
- The promised single routed authority for execution and cache policy is not met for terminal commitment.
- The promised rejection of malformed backend results at their boundary is not met for resident inner exports.
- Device-resident proving and new proving-operation implementations remain explicitly deferred; they are not additional findings.

The operation, resource, and retained-state traits own genuine backend/lifetime boundaries. The corrections above should strengthen those existing authorities rather than add forwarding wrappers or a parallel validation path.

PR #22, opened 2026-09-10
feat(planner): support large-variable schedule searches

RadNi:
## Summary

Add planner support for large-variable schedules whose candidate space may contain paths exceeding the transcript-query capacity.

Oversized candidates are treated as infeasible search paths instead of aborting the entire planner. The planner continues searching for a supported alternative while preserving its original recursive-prefix pruning and budget-independent suffix cache.

## Background

Each fold challenge group contributes one group-root transcript query and one indexed query for every polynomial/live-block coordinate:

`Q_g = 1 + num_polynomials_g × num_live_blocks_g`

The capacity applies to the complete proof schedule across all folds:

`expanded_query_count < u32::MAX`

Large-variable inputs can produce enough live `s_i` blocks that a candidate suffix is valid by itself but exceeds the limit after its parent folds are included.

## Planner behavior

Canonical transcript pricing now carries both values needed by schedule search:

```rust
TranscriptGrindingCost {
    total_nonce_bits,
    expanded_query_count,
}
```

The planner accumulates the exact query contribution of every selected fold. Because an edge's query count depends on its successor, the root search carries finalized prefix queries together with one pending incoming edge. That edge is finalized as soon as the next fold or terminal is selected.

A complete candidate is classified as follows:

```text
root prefix + incoming edge + suffix queries < u32::MAX  → retain candidate
root prefix + incoming edge + suffix queries ≥ u32::MAX  → skip candidate
arithmetic, geometry, or invariant error                 → return error
```

If every candidate exceeds the capacity, planning returns `UnsupportedSchedule`.

## Memoization and pruning

The primary suffix memo remains independent of the caller's remaining query budget. It stores the same objective-optimal suffix frontiers as before and uses the original proof/setup recursive-prefix pruning strategy.

On a memo lookup:

1. The planner checks the cached suffix against the caller's exact root-prefix query count.
2. If the cached result fits, it is reused immediately.
3. If it does not fit, that suffix state is re-expanded under the caller's query prefix to find a feasible alternative.
4. The caller-specific result is not written back to the primary memo.

This is a single planner search with a lazy constrained fallback, not a separate second planning pass. It also avoids adding the remaining query budget to the memo key or maintaining a frontier for every possible budget.

The pruning strategy remains sound in both modes:

- The primary memo search is budget-independent, so its original objective lower bounds can safely prune objectively worse recursive prefixes.
- During constrained re-expansion, every candidate retained for a state is feasible under the same exact caller prefix, so objective pruning cannot discard a candidate needed only because of a different query budget.

Query count is therefore a feasibility constraint rather than a new optimization objective. Existing canonical schedule ordering remains unchanged.

## Result

- Large-variable searches continue after encountering an oversized candidate.
- Query capacity is checked across the entire root-to-terminal schedule.
- Cached suffixes are reused in the common case.
- An infeasible cached optimum triggers local re-expansion for a viable alternative.
- Recursive-prefix early pruning remains enabled.
- Existing schedule artifacts remain unchanged.
- Genuine arithmetic, geometry, and invariant failures are still reported.

## Testing

Coverage includes:

- Exact acceptance and rejection at the transcript-query boundary.
- Root-prefix, incoming-edge, and cached-suffix query accounting.
- Oversized candidates being skipped while valid alternatives remain selectable.
- Budget-independent cache reuse and constrained fallback behavior.
- Recursive-prefix pruning under both unconstrained and constrained search.
- Optimized planner results matching the exhaustive unpruned oracle.
- Schedule artifact generation with no drift.
- Repository formatting and Clippy checks.


omibo:
### `materialize_candidate_schedule` grew to 9 arguments instead of reusing the new cost struct
- **Where:** `crates/akita-schedules/src/runtime.rs:733-746`
- **What:** The PR adds `cached_nonce_bits: usize` and `cached_expanded_query_count: u64` as two more loose parameters, then suppresses the resulting lint:
  ```rust
  #[allow(clippy::too_many_arguments)]
  pub fn materialize_candidate_schedule(
      cached_total: usize,
      cached_nonce_bits: usize,
      cached_expanded_query_count: u64,
      cached_num_setup_field_elements: usize,
      ...
  ```
  Every call site already holds these two values together as one unit. `crates/akita-schedules/src/generated/walk.rs:1081-1096` destructures a `TranscriptGrindingCost` (the PR's own new struct) into two fields just to pass them separately. `crates/akita-planner/src/planner.rs:764-769` and `crates/akita-planner/src/test/unpruned_search.rs:907-912` do the same from `PackedProofCost::nonce_bits()`/`expanded_query_count()`.
- **Fix:** Change the signature to take one `cached_cost: akita_types::TranscriptGrindingCost` instead of `cached_nonce_bits` + `cached_expanded_query_count`. Update the three call sites to construct or pass the struct directly. Drop the `#[allow(clippy::too_many_arguments)]`.

RadNi:
@omibo Fixed in `8b88c624b`. `materialize_candidate_schedule` now takes a single `CandidateMaterializationCost`, whose `grinding` field is the existing `TranscriptGrindingCost`. Planner call sites pass the aggregate directly, and the separate nonce-bit/query-count getters were removed.

Combining only the two grinding arguments still left the function at 8 arguments, above Clippy's limit of 7, so the new materialization-cost value groups all cached metrics checked at this boundary. This also allowed the `#[allow(clippy::too_many_arguments)]` suppression to be removed entirely. The generated `walk.rs` caller referenced above was deleted on current `main`; all current call sites are updated.

Validation passed: formatting, catalog-gen Clippy with `-D warnings`, and the affected crate tests (699 passed, 2 ignored).

quangvdao:
## PR Review findings

Reviewed base `38dbf031013a2b8f7ff93252e190dac2b3353706` through head `da2386dc3daefaa47e8221d906c3fb10ffe18aa4`.

### 1. Blocker — query-blind pruning can discard the only query-feasible schedule

The PR's primary contract is that, when the objective-optimal suffix exceeds the transcript-query capacity under its caller prefix, constrained re-expansion finds another feasible suffix if one exists.

[`process_candidate_batch`](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/schedule_params/suffix_dp/search.rs#L302-L334) calls `level_candidates` before constructing the query-aware recursive `OpeningSearch`. [`level_candidates`](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/schedule_params/suffix_dp/prune.rs#L8-L64) compares six proof/setup coordinates but does not compare expanded query count.

Scratch instrumentation of the existing `proof_first_uniform_search_matches_unpruned_descriptor` fixture found a concrete pair that passes the pruning equivalence predicate and therefore leads to the same child `SuffixState`:

- retained candidate: 95 exact edge queries, coordinates `[65536, 65536, 32, 256, 256, 0]`;
- discarded candidate: 85 exact edge queries, coordinates `[524288, 327680, 32, 256, 256, 0]`.

For a caller prefix in the resulting ten-query window, the discarded path is below `TRANSCRIPT_GRINDING_QUERY_LIMIT`, while the retained path is above it. Because the lower-query branch is deleted before `QueryPrefix::advance`, the [lazy restricted fallback](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/schedule_params/suffix_dp/search.rs#L366-L395) cannot recover it. The planner can return `UnsupportedSchedule` or miss a feasible optimum even though the PR promises to continue past an oversized candidate.

**Required correction:** make constrained-search state visible to candidate generation. During restricted search, do not apply objective-only reductions before exact prefix feasibility is known. For candidates with a common successor, retain a Pareto frontier that includes exact incoming-edge query cost, or require any dominating candidate to have no greater query cost. Apply objective pruning after feasibility. Audit earlier query-blind reductions, particularly [`FoldCandidatePolicy::Best`](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/schedule_params/suffix_dp/candidates.rs#L646-L659), under the same invariant.

**Regression to add:** drive a restricted recursive state with the reproduced 95-query/85-query alternatives and a prefix in that ten-query window. Assert that the 85-query branch is retained and selected. The test should exercise the production constrained fallback, not only `PackedProofCost` or a synthetic local array.

### 2. Moderate — the “unpruned” oracle shares the same capacity blind spot

The oracle's [`candidate_dominates`](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/test/unpruned_search/frontier.rs#L67-L79) calls `strictly_better_for_every_parent`, which ignores expanded query count. The new unit test explicitly records that [`PackedProofCost` dominance ignores queries](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/test/schedule_params.rs#L87-L93). Capacity is checked only after individual suffix/root costs are prepended, so the oracle can discard a proof-worse, lower-query suffix before a later prefix makes it uniquely feasible.

The optimized search and the purportedly exhaustive oracle can therefore share the same miss and still agree; the current optimized-vs-unpruned tests do not establish the new pruning claim.

**Required correction:** make the oracle genuinely capacity-exhaustive. Retain query tradeoffs until complete root candidates are formed, or key oracle states by remaining query capacity. It may reuse low-level canonical transcript accounting, but it must not reuse production candidate pruning or dominance. Mutation-test it against the 95-versus-85 production defect.

### 3. Moderate — the normative documentation contradicts the new behavior

The active transcript-grinding specification says schedule generation [`MUST fail if a candidate cannot derive a valid plan`](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/specs/transcript-grinding.md#L955-L981). This PR instead classifies a query-oversized candidate as an infeasible path, skips it, and continues. The owning [Book section](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/book/src/how/transcript.md#L214-L256) also does not describe the `< 2^32` capacity, constrained fallback, or all-candidates-infeasible result.

**Required correction:** update the active specification and Book to distinguish an expected query-capacity infeasibility from arithmetic, geometry, and invariant failures. State that oversized paths are skipped, planning succeeds with the objective-best feasible complete schedule, and `UnsupportedSchedule` is returned only when the complete intentional search domain contains no feasible schedule.

## Specification alignment

| Claimed behavior | Assessment |
|---|---|
| Exact complete-schedule query accounting | Met in the accounting/materialization path reviewed. |
| Skip one oversized complete candidate and keep searching | Implemented at final selection. |
| Recover a feasible suffix when the cached optimum exceeds the caller's budget | **Not met:** pre-feasibility pruning can delete that suffix. |
| Recursive pruning remains sound in constrained mode | **Not met:** exact query cost is absent from the dominance relation. |
| Optimized search matches an exhaustive oracle | **Not established:** the oracle shares query-blind dominance. |
| Normative and narrative documentation | **Not met:** the active spec states the opposite candidate-failure behavior and the Book omits the new contract. |

`TranscriptGrindingCost`, `CandidateMaterializationCost`, `PendingQueryEdge`, `QueryPrefix`, and `QuerySearch` own real accounting or state-transition invariants and are earned abstractions. I did not find a separate wrapper/duplication issue in their final form. The design problem is the ordering of feasibility and pruning, not those types themselves.

## Validation status

- Exact current-head diagnostic reproduced the 95-query versus 85-query contradiction.
- Focused transcript-cost, prefix-accounting, planner/oracle, and schedule-materialization checks otherwise passed.
- At the last refresh, 42 GitHub checks had succeeded, 4 were still running, and 1 was skipped. Passing CI does not cover the blocker because the oracle shares its dominance assumption.

quangvdao:
Precision correction to the review above: the current implementation deliberately accepts `expanded_query_count < u32::MAX`, so its maximum accepted count is `2^32 - 2`. The spec and runtime error say “fewer than `2^32`,” which would also permit `2^32 - 1`. Please either document the stricter `< u32::MAX` contract everywhere and fix the error text, or change the exclusive limit to `1u64 << 32` if the spec is intended to govern. This one-entry mismatch does not change Finding 1.

quangvdao:
## Runtime-preserving required implementation

Given that planner runtime is already a first-order constraint, I do **not** recommend a global query-Pareto frontier, a query budget in the memo key, or an unconditional second pass. Those designs would make every successful ordinary search pay for a limit that is almost never approached.

The desired implementation is the current lazy architecture with one correctness repair:

1. Keep the primary `ScheduleMemo` budget-independent and keep the existing objective-only frontiers for `Root`/`Unconstrained` search.
2. Reuse the memoized result immediately when it fits the exact `QueryPrefix`.
3. Only when that cached result does not fit, re-expand that state locally as `Restricted(prefix)`, and never write the restricted result into the primary memo.
4. In that restricted re-expansion only, bypass the query-blind pre-feasibility `level_candidates` prune. All candidates that reach the normal `StateFrontiers` have already passed exact prefix feasibility, so the existing objective dominance and guided lower-bound pruning remain valid and should stay enabled.

Concretely, the minimal correctness change at [`process_candidate_batch`](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/schedule_params/suffix_dp/search.rs#L302-L334) is conceptually:

```rust
let candidates = if is_root_level || matches!(query_search, QuerySearch::Restricted(_)) {
    candidates_with_source
} else {
    prune::level_candidates(opening_layout, candidates_with_source)?
};
```

This addresses the demonstrated defect directly: [`level_candidates`](https://github.com/LayerZero-Labs/akita/blob/da2386dc3daefaa47e8221d906c3fb10ffe18aa4/crates/akita-planner/src/schedule_params/suffix_dp/prune.rs#L8-L64) currently discards a 85-query edge in favor of a 95-query edge before the restricted search checks feasibility. It does **not** disable that prune on the common path.

### Keep the restricted path fast

The cached objective winners should also seed the local restricted `StateFrontiers`, retaining only candidates admitted by the current prefix and preserving their original parent-observable key/projection. They are valid feasible incumbents, so the existing strict objective bounds can prune re-expanded branches against them. This avoids rebuilding the fallback frontier from empty while remaining sound. Use the existing canonical `consider_candidate`/frontier insertion logic rather than adding a parallel dominance implementation.

The invariant is:

> Before exact prefix feasibility is known, discard a candidate only if the prune proves it cannot use more queries for every common successor. After candidates are known feasible under the same `QueryPrefix`, the existing objective-only dominance is sound.

`FoldCandidatePolicy::Best` should remain unchanged for runtime and be documented as the planner's candidate-domain boundary. In other words, `UnsupportedSchedule` means “no feasible schedule in the configured candidate domain,” not “no mathematically possible schedule.” If the intended contract is broader than that, widen `Best` only during restricted fallback and benchmark it; do not widen the normal search globally.

I would specifically avoid:

- adding query count to the primary frontier or objective ordering;
- adding remaining query budget to memo keys;
- running a full second search after every successful unconstrained search;
- globally disabling recursive candidate pruning;
- making the production query limit configurable just to test this path.

### Acceptance tests and runtime gate

- Add the reproduced same-child-state regression: the objective-dominating edge uses 95 queries, the discarded alternative uses 85, and a prefix in the 10-query gap must select the 85-query schedule.
- Correct the “unpruned” oracle so it does not share the production `level_candidates` prune.
- Assert the unrestricted selected descriptors and generated schedule artifacts are unchanged.
- Add a cache-order test with two prefixes and prove restricted results are not inserted into the primary memo.
- Benchmark the actual production lookup key/policy that motivated this feature. On current head `da2386d`, my direct OneHot scalar probes for `nv = 40, 44, 48, 52, 56, 60` selected 17,204; 66,829; 264,206; 1,052,116; 4,201,193; and 8,404,588 expanded queries respectively, never entering the fallback and remaining far below the limit. If there is no real input that approaches the limit, the runtime-optimal decision is to defer/remove this feature rather than carry unmotivated complexity.

Finally, align the prose with the actual strict check: `expanded_query_count < u32::MAX` permits at most `2^32 - 2`, not `2^32 - 1`.

crates/akita-planner/src/schedule_params/suffix_dp/search.rs
```diff
@@ -286,7 +304,7 @@ fn process_candidate_batch(
     // Complete-root candidates are traversed in exact lower-bound order.
     // Recursive states do not have that global admission rule and retain
     // local Pareto pruning.
-    let candidates = if is_root_level {
+    let candidates = if is_root_level || matches!(query_search, QuerySearch::Restricted(_)) {
         candidates_with_source
     } else {
         prune::level_candidates(opening_layout, candidates_with_source)?
```
sumchecker, line 310: [P2] Include setup-prefix candidate generation in the restricted-pruning audit

This bypass repairs `level_candidates`, but setup-prefixed states arrive here after `prepare_recursive_level_search` has called the cached `derive_setup_prefix_groups`. That earlier frontier compares only physical witness width and padded setup size; restricted mode is not passed to it.

I reproduced a query tradeoff in the existing `setup_prefix_cache_separates_equal_width_opening_methods` fixture: for `n_prefix=16384`, `d_A=128`, `d_B=64`, opening basis 3, and EvaluationTrace, `[396288, 131072]` with 16 live blocks dominates `[396288, 262144]` with 8 live blocks. The prefix group's query contributions are 17 versus 9. The final frontier also discards valid groups with fewer than eight blocks. Re-expanding this state currently reuses the same query-blind prefix cache.

Please make this reduction query-safe during restricted search, or explicitly justify it as an intentional candidate-domain restriction and test that contract. Preserve the ordinary cache/pruning path. This probe establishes the local loss of lower-query choices; I have not reproduced a complete production root lookup that fails because of it, so the end-to-end impact still needs a constrained setup-offloading regression.

crates/akita-planner/src/schedule_params/suffix_dp/search.rs
```diff
@@ -324,6 +343,16 @@ pub(crate) fn derive_selected_suffix_schedule(
     memo: &mut ScheduleMemo,
     state: SuffixState,
     depth: usize,
+) -> Result<Arc<SuffixResult>, AkitaError> {
+    derive_selected_suffix_schedule_with_query_search(ctx, memo, state, depth, QuerySearch::Root)
```
sumchecker: [P3] Remove the pass-through search entry point

This function only forwards the same arguments and adds `QuerySearch::Root`. The repository explicitly requires one canonical function per concept and forbids pass-through wrappers. Have the sole planner caller pass the root mode to the canonical search function directly, and remove this forwarding entry point. This removes an extra name and call layer without changing search behavior.

crates/akita-planner/src/schedule_params/suffix_dp/search.rs
```diff
@@ -334,7 +363,37 @@ pub(crate) fn derive_selected_suffix_schedule(
...
+                    QuerySearch::Unconstrained,
+                )
+            },
+            Ok,
+        )?;
+        let QuerySearch::Restricted(prefix) = &query_search else {
+            return Err(AkitaError::InvalidSetup(
+                "restricted suffix search lost its query prefix".into(),
+            ));
+        };
```
sumchecker: [P3] Match the restricted mode once and remove this impossible error

The enclosing branch already matched `QuerySearch::Restricted(_)` at line 369, and nothing between that check and this destructuring mutates `query_search`. This error arm cannot execute. Bind `prefix` in the enclosing condition with `if let QuerySearch::Restricted(prefix) = &query_search`, then delete this second check and its `InvalidSetup` error.

crates/akita-planner/src/schedule_params/suffix_dp/search.rs
RadNi: Addressed in e3c5a83d9 using the intentional candidate-domain option from the review. `derive_setup_prefix_groups` now documents that physical-width/setup Pareto pruning defines the setup-prefix domain and that restricted query search intentionally reuses it without adding query-only tradeoffs. The existing reproduced fixture now pins that contract for `n_prefix=16384`, `d_A=128`, `d_B=64`, opening basis 3, EvaluationTrace, and 4 chunks: the 16-live-block layout remains in the domain and the 8-live-block layout is excluded. The ordinary cache/pruning path is unchanged.

crates/akita-planner/src/schedule_params/suffix_dp/search.rs
RadNi: Addressed in e3c5a83d9. I removed the pass-through entry point and made `derive_selected_suffix_schedule` the single canonical function. The sole planner caller now passes `QuerySearch::Root` directly, and recursive callers use the same function.

crates/akita-planner/src/schedule_params/suffix_dp/search.rs
RadNi: Addressed in e3c5a83d9. The enclosing branch now uses `if let QuerySearch::Restricted(prefix) = &query_search`, and the redundant second match plus impossible `InvalidSetup` error were removed.

quangvdao:
## Follow-up re-review (`e3c5a83`)

The original correctness blocker is fixed in the intended runtime-conscious way: ordinary planning keeps the existing pruning, while `Restricted` search bypasses the unsafe query-blind frontier pruning only after the cached schedule violates the transcript bound.

My revised assessment has two remaining improvements:

### 1. Add one production-path fallback regression

The new tests validate query-aware dominance and prefix accounting independently, but none drives the actual production sequence:

```text
cached optimum exceeds the limit
→ cache is rejected
→ Restricted fallback runs
→ a lower-query alternative survives
→ the planner returns it
```

One end-to-end planner fixture covering that sequence is sufficient. It should use a test-only geometry or mechanism and should not make the production query limit configurable. This protects the exact integration boundary where the original defect occurred, without changing normal-path runtime.

The additional cache-order and all-infeasible cases would be useful hardening, but I do not think they need to become separate merge requirements if the central production-path regression is strong.

### 2. State the bounded candidate-domain contract explicitly

Keeping setup, fold, terminal, and slice candidate heuristics bounded is a reasonable design given planner runtime constraints. Consequently, however, `UnsupportedSchedule` means that no feasible schedule was found among layouts admitted by those policies; it does not prove that no mathematically valid layout exists outside the discarded candidates.

I suggest replacing the vague phrase “configured candidate domain” with an explicit statement such as:

> Restricted query search considers schedules admitted by the planner's existing bounded candidate-generation policies. `UnsupportedSchedule` means that no feasible schedule was found within this candidate domain; it does not prove that no mathematically valid schedule exists outside layouts discarded by those policies.

This is a documentation/contract clarification, not a request to widen candidate generation.

### Runtime recommendation

I would keep the current architecture:

- no full query-aware Pareto frontiers on the normal path;
- no query budget in primary memo keys;
- no global enumeration of setup-prefix or recursive layouts;
- no production-configurable artificial limit;
- larger exploration only in the rare `Restricted` fallback.

With the production-path regression and the bounded-domain clarification, the implementation would match the intended runtime-preserving design.

RadNi:
@quangvdao Addressed both remaining improvements in `26f25239a`.

1. The existing production/oracle fixture now also exercises the complete fallback sequence through `find_schedule_in_relation_order`: it derives the normal winner’s exact query count, applies a test-only root-prefix offset that makes that winner hit the strict limit, reruns the real planner, and asserts that the returned schedule has a strictly lower query count. The production limit remains fixed and normal-path behavior is unchanged.

2. The Book and active transcript-grinding specification now state explicitly that restricted search considers schedules admitted by the planner’s bounded candidate-generation policies, and that `UnsupportedSchedule` does not prove that no valid layout exists outside candidates discarded by those policies.

Validation passed: the new production-path regression, focused suffix-DP tests, documentation guardrails, Rust/TOML formatting, catalog-gen Clippy, and all three CI Clippy feature graphs with `-D warnings`.

quangvdao:
## Re-review follow-up (`26f25239a`)

The bounded candidate-domain clarification is now resolved in both the Book and active specification, and I found no production accounting or runtime defect in the test hook. One test-coverage issue remains.

### Moderate — the fallback regression does not detect the original pruning bug

The new production-path test correctly demonstrates that the cached winner is rejected at the strict limit and that the planner returns some lower-query schedule. However, it only asserts:

```text
constrained_query_count < unconstrained_query_count
```

It does not require the restricted search to recover the reproduced lower-query branch or even retain the same root.

I mutation-tested the exact repair in an isolated checkout:

1. The test passes unchanged on `26f25239a`.
2. I restored the old query-blind behavior by changing the candidate condition from
   `is_root_level || Restricted` back to `is_root_level`.
3. The test still passed.
4. Separately, forcing unconditional cache reuse made the test fail.

So the test covers cache rejection, but it does not protect the Restricted-mode pruning fix. Another lower-query schedule can win and mask restoration of the original bug.

**Suggested correction:** constrain the fixture to the known same-successor 95-query/85-query pair, place the prefix inside their ten-query feasibility gap, and assert that the exact 85-query edge or descriptor is selected. Alternatively, directly exercise that suffix state and memo, asserting cached rejection, restricted selection, and that the restricted result does not replace the unconstrained cached result.

The test-only prefix seam remains excluded from production builds, and I found no material normal-path overhead. At the latest refresh, 42 checks had passed, 4 were pending, 1 was skipped, and none had failed.

RadNi:
@quangvdao Addressed in `0e3fcb1bf`.

The new focused regression drives the production restricted-batch path with only the reproduced same-successor candidates. It asserts:

- both candidates resolve to the same child memo key;
- their exact edge query counts are `95` and `85`;
- the prefix is placed using the 85-query candidate's feasible budget;
- the exact 85-query descriptor survives the restricted frontier; and
- restricted re-expansion does not replace the unconstrained memo entry (`Arc::ptr_eq`).

I also mutation-tested the reviewed condition: changing `is_root_level || Restricted` back to `is_root_level` makes this test fail because the 85-query descriptor is pruned; restoring the fix makes it pass.

Validation is green for the focused/catalog tests, Rustfmt, Taplo, the Rust line-cap gate, catalog-gen Clippy, and all three CI Clippy feature configurations.

sumchecker (approved):

PR #23, opened 2026-09-11
refactor(prover): introduce staged executor

omibo:
## Summary

- Introduce a stack-only `ProverExecutor` for admission, resource preparation, root proving, and suffix proving.
- Keep the transcript, setup, prefix slots, schedule, and proof progress explicit.
- Split nonterminal proving into named commitment, compression, range-image, and sumcheck stages with named outputs.
- Remove optional successor plumbing, `FinishFoldArgs`, and the late claim-binding callback.

This is an orchestration refactor. It preserves protocol behavior, transcript order, arithmetic kernels, large-buffer ownership, and accepted witness reuse.

## Validation

- 443 prover tests passed; 3 skipped
- 39 focused PCS tests passed
- 1,827 CI-profile tests passed
- All required Clippy feature graphs and repository guardrails passed


quangvdao:
## PR Review findings

Reviewed exact active surface: `a5ab55cc7e816832afe041ea089caa6399d2485e...0d67f4fd34921b52d4b25b084c847f8ae39e827e`.

PR #20 merged while this review was being published. Findings 1–3 were introduced by that now-merged base rather than by PR #23, but they remain present in PR #23's active tree and are consolidated here so the open stack has one complete review thread. Finding 4 is specific to PR #23.

### 1. Major — caller-supplied short-norm extrema can select an invalid commitment kernel

[`ShortNormRepresentation::new`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/commitment/source.rs#L295-L341) forwards caller-supplied `negative_abs_max` and `positive_max` into `PackedSignedDigitView`. [`from_encoded`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/backend/packed_digits/mod.rs#L403-L440) checks that those claimed extrema fit the stored bit width, but never compares them with the decoded coefficients. The selected-representation compiler also checks only extents and non-emptiness ([`compiler.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/commitment/source/compiler.rs#L109-L117)).

**Concrete failure:** a downstream `CommitmentSource` can materialize encoded bytes `[0x03]`, live length `1`, physical length `64`, signed width `3`, and claimed extrema `(0, 1)`. Construction succeeds although the first decoded coefficient is `+3`. Under a one-digit plan with `log_basis_inner = 2`, the CPU path trusts the false bounds and selects the balanced kernel ([`commitment.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/compute/cpu/commitment.rs#L420-L465)). Debug builds hit the active-range assertion. On LUT-backed release paths, masking maps `+3` to the entry for `-1`: `(3 + 2) & (4 - 1) = 1` ([`lut.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-algebra/src/ring/crt_ntt_repr/lut.rs#L150-L165)), producing a commitment to different coefficients. Accelerated release paths may process `+3` directly, making behavior backend/path-dependent.

**Required correction:** derive the exact extrema by scanning the live encoded coefficients during selected-representation materialization, or reject when derived and supplied extrema differ. Only verified extrema should control balanced/raw dispatch. Add a regression using the payload above, including scalar release parity.

### 2. Moderate — terminal `InnerOnly` routing is lost before cache-policy evaluation

`NttExecutionRequirements::add_terminal` correctly emits an `InnerOnly` request ([`requirements.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/compute/requirements.rs#L477-L503)). `CommitmentExecutor::retained_routed_requirement` uses that route to select split-inner resources, but then converts it to `CommitmentNttRequirement`, which has no route field. [`CommitmentNttRequirement::routed`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/commitment/resources.rs#L66-L116) reconstructs every request as `InnerOuter` before calling the backend's cache policy ([`resources.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/commitment/resources.rs#L230-L250)).

A backend permitted to stream terminal A while caching nonterminal A/B will therefore classify the terminal operation as cached and prewarm/build a slot that runtime policy intended to stream. This contradicts the active spec's requirement that one resolved route control prewarm, owner selection, and execution.

**Required correction:** carry `CommitmentNttRoute` in `CommitmentNttRequirement`, or pass the original `RoutedNttRequirement` unchanged through `CommitmentResourceControl`. Add a route-sensitive backend test asserting that terminal retention and prewarm both observe `InnerOnly`.

### 3. Moderate — resident inner exports are not fully validated or frozen at their boundary

[`ResidentCommitmentState::inner_relation_material`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/commitment/state_policy.rs#L512-L545) invokes the backend exporter and then constructs `InnerRelationStateMaterial`. That constructor checks only that each row is divisible by the ring dimension ([`state_policy.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/commitment/state_policy.rs#L341-L359)); it does not enforce `rows.len() == binding.source_count()` or the exact per-row length `num_live_blocks * n_a * ring_dimension`.

For example, with `D=64`, `num_live_blocks=1`, `n_a=2`, and one source, an exporter can return one 64-coefficient row. It passes the boundary constructor although 128 coefficients are required. The exact length is rejected only later during ring-switch construction ([`coeffs.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/protocol/ring_switch/coeffs.rs#L397-L419)), after transcript-dependent work.

There is also a latent consistency failure for stateful custom exporters: the terminal suffix exports once to construct and absorb `terminal_t_fields_message`, then exports again to obtain the proof row ([`suffix.rs`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/protocol/core/suffix.rs#L270-L276)). Built-in CPU exporters are pure, but the public trait does not enforce purity; alternating same-shaped rows would bind row A and prove with row B.

**Required correction:** introduce one canonical constructor/helper that takes the binding and exported rows, checks ring dimension, source count, and exact checked coefficient lengths, and returns frozen material. Terminal binding and relation construction should derive from the same validated export.

### 4. Low — the staged-executor refactor leaves live documentation naming deleted helpers

PR #23 replaces `RingRelationProver::new` with [`RingRelationProver::prepare`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/protocol/ring_relation.rs#L375-L413) and replaces `finish_prepared_fold` with [`prepare_fold_relation`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-prover/src/protocol/core/fold/mod.rs#L61-L75), but current contributor-facing sources still direct readers to the deleted names:

- the canonical Book names [`RingRelationProver::new`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/book/src/how/proving/akita-fold-realizations.md#L902-L929);
- the live runtime-ring spec names [`RingRelationProver::new`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/specs/runtime-ring-cutover.md#L584-L594);
- verifier root-replay documentation says it mirrors [`finish_prepared_fold`](https://github.com/LayerZero-Labs/akita/blob/0d67f4fd34921b52d4b25b084c847f8ae39e827e/crates/akita-verifier/src/protocol/core/root_fold.rs#L129-L140).

Neither deleted symbol exists at this head.

**Required correction:** update the references to `RingRelationProver::prepare` / `prepare_fold_relation`, and extend the applicable documentation dead-symbol guardrail so the same drift is caught in future renames.

## Specification and abstraction assessment

- PR #23 itself preserves transcript sequencing, next-witness binding, ring-switch finalization, stage order, setup-prefix ownership, terminal separation, and accepted-witness reuse. I found no additional protocol-correctness issue in its ten changed prover files.
- The independent commitment stages and backend-owned state are earned abstractions. The inherited gaps above should be fixed by strengthening their existing route and state authorities, not by adding forwarding wrappers or a parallel validation path.
- PR #23 is now based directly on `main` after #20's merge, and fresh workflows were triggered for this exact surface. They were still running at the time of this comment.

RadNi (approved):

PR #26, opened 2026-09-11
fix(prover): harden commitment backend boundaries

omibo:
Three commitment extension boundaries accepted inconsistent or incomplete backend data. A custom short-norm source could understate its decoded extrema and select the wrong release kernel, terminal NTT requirements lost their `InnerOnly` route before backend cache-policy evaluation, and a resident inner exporter could return malformed or different rows across terminal transcript binding and relation construction.

This change:

- derives exact extrema from the live packed short-norm coefficients and rejects mismatched caller-supplied bounds before kernel dispatch;
- carries `CommitmentNttRoute` through `CommitmentNttRequirement`, including cache-policy and prewarm calls;
- validates exported resident inner rows against the complete commitment binding, including ring dimension, source count, and exact per-source length;
- freezes the first validated resident export so transcript binding and relation construction use identical material, and includes that material in retained-byte accounting; and
- updates the composable commitment execution specification with these contracts.

### DIFF
--- a/crates/akita-prover/src/backend/field_reduction.rs
+++ b/crates/akita-prover/src/backend/field_reduction.rs
@@ -71,4 +71,30 @@ mod tests {
         assert_eq!(packed.to_i8_digits().len(), 2 * D);
         assert_eq!(packed.committed_coeff_len().unwrap(), 2 * D);
     }
+
+    #[test]
+    fn tensor_packed_extrema_in_physical_tail_are_admitted() {
+        type F = Prime32Offset99;
+        type E = FpExt4<F>;
+        const D: usize = 128;
+        let logical_len = D + <E as ExtField<F>>::DEGREE;
+        let mut digits = vec![0; logical_len];
+        digits[D + 1] = 3;
+        let witness = RecursiveWitnessFlat::from_i8_digits(digits);
+        let packed = tensor_pack_recursive_witness::<F, E, D>(&witness).unwrap();
+        let stored = packed.to_i8_digits();
+        assert!(stored[logical_len..].contains(&3));
+        let (bytes, width, negative, positive) = packed.packed_representation_parts();
+
+        crate::commitment::ShortNormRepresentation::new(
+            bytes,
+            packed.live_coeff_len(),
+            stored.len(),
+            packed.committed_coeff_len().unwrap(),
+            width,
+            negative,
+            positive,
+        )
+        .unwrap();
+    }
 }
--- a/crates/akita-prover/src/backend/packed_digits/mod.rs
+++ b/crates/akita-prover/src/backend/packed_digits/mod.rs
@@ -429,6 +429,28 @@ impl<'a> PackedSignedDigitView<'a> {
             positive_max,
         };
         validate_bounds(bounds, bit_width)?;
+        let mut decoded_bounds = SignedDigitBounds {
+            negative_abs_max: 0,
+            positive_max: 0,
+        };
+        for index in 0..live_len {
+            let digit = scalar::decode_at_zero_padded(storage, index, bit_width);
+            if digit < 0 {
+                decoded_bounds.negative_abs_max =
+                    decoded_bounds.negative_abs_max.max(digit.unsigned_abs());
+            } else {
+                decoded_bounds.positive_max = decoded_bounds.positive_max.max(digit as u8);
+            }
+        }
+        if decoded_bounds != bounds {
+            return Err(AkitaError::InvalidInput(format!(
+                "packed signed-digit bounds [-{}, {}] disagree with decoded live bounds [-{}, {}]",
+                bounds.negative_abs_max,
+                bounds.positive_max,
+                decoded_bounds.negative_abs_max,
+                decoded_bounds.positive_max,
+            )));
+        }
         Ok(Self {
             storage,
             stored_len: represented_len.min(physical_len),
--- a/crates/akita-prover/src/backend/packed_digits/mod.rs
+++ b/crates/akita-prover/src/backend/packed_digits/mod.rs
@@ -403,26 +403,39 @@ impl<'a> PackedSignedDigitView<'a> {
     pub(crate) fn from_encoded(
         storage: &'a [u8],
         live_len: usize,
+        stored_len: usize,
         physical_len: usize,
         bit_width: u8,
         negative_abs_max: u8,
         positive_max: u8,
     ) -> Result<Self, AkitaError> {
         validate_bit_width(bit_width)?;
-        if live_len == 0 || live_len > physical_len {
+        if live_len == 0 || live_len > stored_len || stored_len > physical_len {
             return Err(AkitaError::InvalidInput(
                 "packed signed-digit view has inconsistent extents".into(),
             ));
         }
-        let represented_bits = checked::product([storage.len(), 8]).ok_or_else(|| {
-            AkitaError::InvalidInput("packed signed-digit represented length overflow".into())
+        let stored_bits =
+            checked::product([stored_len, usize::from(bit_width)]).ok_or_else(|| {
+                AkitaError::InvalidInput("packed signed-digit stored length overflow".into())
+            })?;
+        let expected_bytes = checked::div_ceil(stored_bits, 8).ok_or_else(|| {
+            AkitaError::InvalidInput("packed signed-digit byte length overflow".into())
         })?;
-        let represented_len = represented_bits / usize::from(bit_width);
-        if represented_len < live_len {
+        if storage.len() != expected_bytes {
             return Err(AkitaError::InvalidSize {
-                expected: live_len,
-                actual: represented_len,
+                expected: expected_bytes,
+                actual: storage.len(),
             });
+        }
+        let used_final_bits = stored_bits % 8;
+        if used_final_bits != 0 {
+            let unused_mask = !((1u8 << used_final_bits) - 1);
+            if storage.last().is_some_and(|byte| byte & unused_mask != 0) {
+                return Err(AkitaError::InvalidInput(
+                    "packed signed-digit payload has non-canonical trailing bits".into(),
+                ));
+            }
         }
         let bounds = SignedDigitBounds {
             negative_abs_max,
@@ -433,7 +446,7 @@ impl<'a> PackedSignedDigitView<'a> {
             negative_abs_max: 0,
             positive_max: 0,
         };
-        for index in 0..live_len {
+        for index in 0..stored_len {
             let digit = scalar::decode_at_zero_padded(storage, index, bit_width);
             if digit < 0 {
                 decoded_bounds.negative_abs_max =
@@ -444,7 +457,7 @@ impl<'a> PackedSignedDigitView<'a> {
         }
         if decoded_bounds != bounds {
             return Err(AkitaError::InvalidInput(format!(
-                "packed signed-digit bounds [-{}, {}] disagree with decoded live bounds [-{}, {}]",
+                "packed signed-digit bounds [-{}, {}] disagree with decoded stored bounds [-{}, {}]",
                 bounds.negative_abs_max,
                 bounds.positive_max,
                 decoded_bounds.negative_abs_max,
@@ -453,7 +466,7 @@ impl<'a> PackedSignedDigitView<'a> {
         }
         Ok(Self {
             storage,
-            stored_len: represented_len.min(physical_len),
+            stored_len,
             bit_width,
             bounds,
             vector_safe: false,
--- a/crates/akita-prover/src/backend/packed_digits/tests.rs
+++ b/crates/akita-prover/src/backend/packed_digits/tests.rs
@@ -294,6 +294,7 @@ fn borrowed_encoded_view_matches_owned_view_without_load_padding() {
     let borrowed = PackedSignedDigitView::from_encoded(
         packed.encoded_bytes(),
         digits.len(),
+        digits.len(),
         256,
         packed.bit_width(),
         bounds.negative_abs_max(),
--- a/crates/akita-prover/src/commitment/resources.rs
+++ b/crates/akita-prover/src/commitment/resources.rs
@@ -66,6 +66,7 @@ pub enum CommitmentNttRoute {
 /// Exact cache request routed to one registered commitment stage.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct CommitmentNttRequirement {
+    route: CommitmentNttRoute,
     stage: CommitmentNttStage,
     key: NttCacheKey,
     routing_extent: usize,
@@ -74,6 +75,7 @@ pub struct CommitmentNttRequirement {
 impl CommitmentNttRequirement {
     /// Construct after validating the operation-level routing extent.
     pub fn new(
+        route: CommitmentNttRoute,
         stage: CommitmentNttStage,
         key: NttCacheKey,
         routing_extent: usize,
@@ -84,10 +86,16 @@ impl CommitmentNttRequirement {
             ));
         }
         Ok(Self {
+            route,
             stage,
             key,
             routing_extent,
         })
+    }
+
+    /// Execution route whose cache policy owns this request.
+    pub const fn route(&self) -> CommitmentNttRoute {
+        self.route
     }
 
     /// Owning commitment stage.
@@ -110,7 +118,7 @@ impl CommitmentNttRequirement {
             fold_level: 0,
             cluster: NttOperationCluster::Commit,
             commitment_stage: Some(self.stage),
-            commitment_route: Some(CommitmentNttRoute::InnerOuter),
+            commitment_route: Some(self.route),
             key: self.key,
             routing_extent: self.routing_extent,
         }
@@ -144,7 +152,13 @@ impl CommitmentExecutionPlan {
         let key = NttCacheKey::from_matrix_shape(plan.ring_dimension, plan.n_a, width, domain)?;
         let routing_extent = checked::product([plan.n_a, width])
             .ok_or_else(|| AkitaError::InvalidSetup("commitment A extent overflow".into()))?;
-        CommitmentNttRequirement::new(CommitmentNttStage::Inner, key, routing_extent).map(Some)
+        CommitmentNttRequirement::new(
+            CommitmentNttRoute::InnerOuter,
+            CommitmentNttStage::Inner,
+            key,
+            routing_extent,
+        )
+        .map(Some)
     }
 
     /// Exact B-matrix cache request when this route contains an outer stage.
@@ -161,7 +175,13 @@ impl CommitmentExecutionPlan {
         )?;
         let routing_extent = checked::product([plan.n_b(), width])
             .ok_or_else(|| AkitaError::InvalidSetup("commitment B extent overflow".into()))?;
-        CommitmentNttRequirement::new(CommitmentNttStage::Outer, key, routing_extent).map(Some)
+        CommitmentNttRequirement::new(
+            CommitmentNttRoute::InnerOuter,
+            CommitmentNttStage::Outer,
+            key,
+            routing_extent,
+        )
+        .map(Some)
     }
 }
 
@@ -370,10 +390,13 @@ where
     fn routed_requirement(
         requirement: RoutedNttRequirement,
     ) -> Result<CommitmentNttRequirement, AkitaError> {
+        let route = requirement.commitment_route.ok_or_else(|| {
+            AkitaError::InvalidSetup("commitment NTT requirement has no route discriminator".into())
+        })?;
         let stage = requirement.commitment_stage.ok_or_else(|| {
             AkitaError::InvalidSetup("commitment NTT requirement has no stage discriminator".into())
         })?;
-        CommitmentNttRequirement::new(stage, requirement.key, requirement.routing_extent)
+        CommitmentNttRequirement::new(route, stage, requirement.key, requirement.routing_extent)
     }
 
     fn routed_resources(
@@ -414,11 +437,8 @@ where
         &self,
         requirement: RoutedNttRequirement,
     ) -> Result<(), AkitaError> {
-        let route = requirement.commitment_route.ok_or_else(|| {
-            AkitaError::InvalidSetup("commitment NTT requirement has no route discriminator".into())
-        })?;
         let requirement = Self::routed_requirement(requirement)?;
-        self.routed_resources(route, requirement.stage())?
+        self.routed_resources(requirement.route(), requirement.stage())?
             .ensure_ntt_slot(requirement)
     }
 
@@ -426,11 +446,8 @@ where
         &self,
         requirement: RoutedNttRequirement,
     ) -> Result<Option<NttCacheOwnerId>, AkitaError> {
-        let route = requirement.commitment_route.ok_or_else(|| {
-            AkitaError::InvalidSetup("commitment NTT requirement has no route discriminator".into())
-        })?;
         let requirement = Self::routed_requirement(requirement)?;
-        let resources = self.routed_resources(route, requirement.stage())?;
+        let resources = self.routed_resources(requirement.route(), requirement.stage())?;
         if !resources.requirement_is_cached(requirement)? {
             return Ok(None);
         }
@@ -462,8 +479,20 @@ mod tests {
     #[test]
     fn commitment_requirement_rejects_short_routing_extent() {
         let key = NttCacheKey::from_matrix_shape(64, 2, 8, NttTransformDomain::Negacyclic).unwrap();
-        assert!(CommitmentNttRequirement::new(CommitmentNttStage::Inner, key, 15).is_err());
-        assert!(CommitmentNttRequirement::new(CommitmentNttStage::Inner, key, 16).is_ok());
+        assert!(CommitmentNttRequirement::new(
+            CommitmentNttRoute::InnerOuter,
+            CommitmentNttStage::Inner,
+            key,
+            15,
+        )
+        .is_err());
+        assert!(CommitmentNttRequirement::new(
+            CommitmentNttRoute::InnerOuter,
+            CommitmentNttStage::Inner,
+            key,
+            16,
+        )
+        .is_ok());
     }
 
     #[test]
--- a/crates/akita-prover/src/commitment/resources.rs
+++ b/crates/akita-prover/src/commitment/resources.rs
@@ -66,6 +66,7 @@ pub enum CommitmentNttRoute {
 /// Exact cache request routed to one registered commitment stage.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct CommitmentNttRequirement {
+    fold_level: usize,
     route: CommitmentNttRoute,
     stage: CommitmentNttStage,
     key: NttCacheKey,
@@ -75,6 +76,7 @@ pub struct CommitmentNttRequirement {
 impl CommitmentNttRequirement {
     /// Construct after validating the operation-level routing extent.
     pub fn new(
+        fold_level: usize,
         route: CommitmentNttRoute,
         stage: CommitmentNttStage,
         key: NttCacheKey,
@@ -86,11 +88,17 @@ impl CommitmentNttRequirement {
             ));
         }
         Ok(Self {
+            fold_level,
             route,
             stage,
             key,
             routing_extent,
         })
+    }
+
+    /// Fold level whose compute stack owns this request.
+    pub const fn fold_level(&self) -> usize {
+        self.fold_level
     }
 
     /// Execution route whose cache policy owns this request.
@@ -115,7 +123,7 @@ impl CommitmentNttRequirement {
 
     fn routed(&self) -> RoutedNttRequirement {
         RoutedNttRequirement {
-            fold_level: 0,
+            fold_level: self.fold_level,
             cluster: NttOperationCluster::Commit,
             commitment_stage: Some(self.stage),
             commitment_route: Some(self.route),
@@ -152,13 +160,14 @@ impl CommitmentExecutionPlan {
         let key = NttCacheKey::from_matrix_shape(plan.ring_dimension, plan.n_a, width, domain)?;
         let routing_extent = checked::product([plan.n_a, width])
             .ok_or_else(|| AkitaError::InvalidSetup("commitment A extent overflow".into()))?;
-        CommitmentNttRequirement::new(
-            CommitmentNttRoute::InnerOuter,
-            CommitmentNttStage::Inner,
-            key,
-            routing_extent,
-        )
-        .map(Some)
+        let route = match self.mode() {
+            super::CommitmentExecutionMode::InnerOnly => CommitmentNttRoute::InnerOnly,
+            super::CommitmentExecutionMode::Full | super::CommitmentExecutionMode::Uncompressed => {
+                CommitmentNttRoute::InnerOuter
+            }
+        };
+        CommitmentNttRequirement::new(0, route, CommitmentNttStage::Inner, key, routing_extent)
+            .map(Some)
     }
 
     /// Exact B-matrix cache request when this route contains an outer stage.
@@ -176,6 +185,7 @@ impl CommitmentExecutionPlan {
         let routing_extent = checked::product([plan.n_b(), width])
             .ok_or_else(|| AkitaError::InvalidSetup("commitment B extent overflow".into()))?;
         CommitmentNttRequirement::new(
+            0,
             CommitmentNttRoute::InnerOuter,
             CommitmentNttStage::Outer,
             key,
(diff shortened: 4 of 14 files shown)


quangvdao:
## PR Review findings

Reviewed base `a5ab55cc7e816832afe041ea089caa6399d2485e` through head `d44f72b926348d18635cdf4ca784ebeb28749cda`.

### 1. Major — extrema validation still ignores coefficients consumed from encoded tail bytes

[`PackedSignedDigitView::from_encoded`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/backend/packed_digits/mod.rs#L417-L461) derives extrema over `0..live_len`, but then sets `stored_len` to `represented_len.min(physical_len)`. The commitment decoder can therefore consume complete digits which were not covered by the validated bounds.

A concrete public-input counterexample is:

```rust
ShortNormRepresentation::new(&[0x19], 1, 64, 3, 0, 1)
```

This succeeds: digit 0 is `+1`, matching the claimed bound, while bits 3–5 encode a second complete digit `+3`. One byte represents two complete three-bit digits, so the returned view records bounds of `+1` but decodes `+3` at index 1. I reproduced this at the reviewed head with a focused scratch test.

The CPU commitment path chooses balanced versus raw conversion from the incomplete bounds and then decodes the physical ring containing the hidden coefficient ([dispatch](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/compute/cpu/commitment.rs#L420-L467)). For `num_digits_inner = 1` and `log_basis_inner = 2`, debug kernels reach the digit-range assertion; scalar release conversion masks `+3` into a different value ([LUT](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-algebra/src/ring/crt_ntt_repr/lut.rs#L150-L165)), while accelerated conversion may process `+3` directly. The original debug-panic/release-miscommit and backend-divergence failure therefore remains reachable.

This also creates an honest-input regression: tensor packing preserves a shorter logical `live_coeff_len` while producing a longer physical packed vector ([construction](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/backend/field_reduction.rs#L23-L35)). Built-in bounds cover all stored digits, but [`represent_as`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/source/builtins.rs#L216-L229) now compares them only against the logical-live prefix before replacing the view. Valid tensor material is rejected when its extrema occur only in the physical tail.

**Required correction:** represent the exact stored-coefficient extent explicitly, require or canonicalize unused final-byte bits, and derive bounds over every coefficient commitment decoding can consume. Add the `[0x19]` case through actual CPU commitment, debug/release and scalar/accelerated parity tests, and a valid tensor-packed case whose extrema occur past `live_coeff_len`.

### 2. Moderate — direct terminal prewarming still reports `InnerOuter`

The proof-wide route is now preserved, but [`CommitmentExecutionPlan::inner_ntt_requirement`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/resources.rs#L133-L161) hardcodes `CommitmentNttRoute::InnerOuter` for every plan, including [`for_terminal`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/plan.rs#L181-L207) plans.

Public [`CommitmentExecutor::prewarm_request`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/executor.rs#L346-L385) derives its requirement through that helper, and [`PreparedStage::ensure_ntt_slot`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/prepared.rs#L47-L59) presents the resulting `InnerOuter` value to route-sensitive cache policy. A backend that caches terminal `InnerOnly` but streams `InnerOuter` skips the requested terminal prewarm; the reverse policy creates an unintended slot.

The new route test manually injects `InnerOnly` into the proof-wide methods and exercises direct prewarm only with a full plan, so it misses this path.

**Required correction:** derive the inner requirement route from `plan.mode()`, using `InnerOnly` for terminal plans and `InnerOuter` for full/uncompressed plans. Add route-sensitive direct-terminal-prewarm regressions for both split-inner and fused executors.

### 3. Moderate — proof-wide cache-policy conversion still loses the owning fold level

[`RoutedNttRequirement`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/compute/requirements.rs#L24-L44) carries the owning fold, and terminal planning supplies its actual level. Conversion into `CommitmentNttRequirement` drops that field, while [`routed()` reconstructs every request with `fold_level: 0`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/resources.rs#L116-L124).

A level-sensitive backend therefore sees every proof-wide commitment request as fold 0 and can choose the wrong cache/stream policy. The new tests use fold level 0 only.

**Required correction:** retain `fold_level` in the commitment requirement or pass the original routed requirement unchanged to cache policy. Add a nonzero-fold policy regression.

### 4. Moderate — freezing adds the witness-sized copy forbidden by the active performance contract

`InnerRelationStateMaterial` derives `Clone` over `Vec<RingVec<F>>` ([definition](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/state_policy.rs#L357-L362)), and `RingVec` is `Vec<F>`-backed, so this is a deep clone ([container](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-types/src/proof/containers.rs#L14-L18)).

For the CPU route, the backend state already owns W bytes of inner rows; borrowed export deep-clones those rows, [`inner_relation_material`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/state_policy.rs#L574-L595) deep-clones the material again into the frozen cache, and each cache hit returns another W-byte clone. After first access the resident state retains approximately `2W`, and a live returned material can take the peak to approximately `3W`.

The new retained-byte accounting reports the cached W correctly, but that confirms the regression. The active specification says that no route may add a persistent witness-sized copy beyond the inner image needed by later proving ([performance contract](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/specs/composable-commitment-execution.md#L374-L389)).

The same-head [profile report](https://github.com/LayerZero-Labs/akita/pull/26#issuecomment-5641971287) now provides observable impact: all 13 proving medians regress, ranging from `+12.4%` to `+27.1%`, although every profile still verifies. This is well beyond the specification's unexplained `3%` blocking threshold. The report does not isolate the new cloning as the sole cause, but it confirms that the PR cannot presently satisfy its performance acceptance contract.

**Required correction:** use one shared immutable material allocation and replace/drop the duplicated backend image when freezing, or redesign the handoff so transcript binding and relation construction share or consume a single allocation. Add retained-byte and peak-copy regression coverage.

### 5. Moderate — portable resident exports bypass the validated frozen authority

The repaired [`inner_relation_material`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/state_policy.rs#L574-L595) path locks, binding-validates, and freezes the first export. However, [`portable_hint` and `into_portable_hint`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/state_policy.rs#L701-L716) invoke separate export routines directly. Those routines do not use the frozen rows and assemble the hint without validating the binding's source count or exact planned row length ([export path](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/state_policy.rs#L132-L193)).

A stateful exporter can return valid row A for terminal transcript binding, then return same-shaped row B through `portable_hint`; an undersized row which is merely ring-divisible also passes portable export.

**Required correction:** route every resident inner export, including borrowed and consuming portable export, through one binding-validated authority. Add alternating-exporter and malformed-portable-export tests.

### 6. Moderate — the public material boundary still permits incomplete canonical state

[`InnerRelationStateMaterial::new`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/state_policy.rs#L364-L402) remains public and accepts any ring-divisible rows without the expected source count or exact row length. The public [`InnerRelationState`](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/commitment/state_policy.rs#L420-L432) preflight defaults to success, and its materialization method receives no expected plan.

A custom state can return one D=64 row containing 64 coefficients when the relation requires 128. The prover obtains this material before transcript work, but exact length is rejected only later during ring-switch construction ([late check](https://github.com/LayerZero-Labs/akita/blob/d44f72b926348d18635cdf4ca784ebeb28749cda/crates/akita-prover/src/protocol/ring_switch/coeffs.rs#L397-L419)). This leaves the generic public extension boundary outside the new reject-before-transcript guarantee.

**Required correction:** make binding-aware construction the enforceable public path, or pass the expected plan and source count into generic materialization. Add a custom-state test proving rejection before transcript mutation.

## Specification alignment

| Requirement | Assessment |
|---|---|
| Exact short-norm extrema before kernel dispatch | **Not met:** decoded tail coefficients remain outside validation |
| Preserve terminal route through cache policy and prewarming | Partially met: proof-wide routing is fixed; direct terminal prewarm remains mislabeled |
| Preserve complete routed cache-policy identity | **Not met:** `fold_level` is reconstructed as zero |
| Validate resident rows against the complete binding | Met for `ResidentCommitmentState::inner_relation_material`; not for portable or generic state paths |
| Freeze material shared by terminal transcript and relation | Met for the core resident relation path |
| No persistent extra witness-sized copy | **Not met:** the frozen cache introduces one persistent deep copy |

`CommitmentNttRoute` and binding-aware `InnerRelationStateMaterial::from_binding` are earned abstractions: they carry real routing and shape invariants. The remaining problem is that those invariants are not yet owned by one canonical boundary. Route identity is reconstructed in two forms, while resident rows can still be obtained through frozen, portable, and generic paths with different validation and ownership behavior. The correction should consolidate those authorities rather than add another wrapper.

## Validation

- Reviewed the complete seven-file base-to-head diff at the SHA above.
- Reproduced the packed-tail counterexample with a focused temporary unit test: the constructor succeeds with bounds `+1`, while the same view decodes `+3` at the next stored index.
- The core mutex/frozen-material path is race-safe and does make terminal transcript and relation construction share identical successful material.
- Full/uncompressed and valid fused/split stage ownership appear correct outside the direct terminal-prewarm path.
- Current-head unit tests, Clippy, security, portability, and all seven benchmark shards passed. Schedule artifact drift and both transcript-semantics jobs were still pending at the final snapshot. Passing existing tests does not exercise the counterexamples above.
- The current-head benchmark report verified all 13 profiles, but reported proving-time regressions of `+12.4%` through `+27.1%` across every profile.

omibo:
## Resolution of the six review findings

Thanks for the detailed review. All six behavioral issues are addressed on the current branch:

1. **Stored packed tail and extrema validation**

   `ShortNormRepresentation` now carries `stored_coefficient_len` explicitly. `PackedSignedDigitView::from_encoded` validates the extrema over every stored coefficient that commitment may decode and rejects nonzero unused bits in the final byte. Akita-owned immutable `PackedSignedDigits` reuse the exact bounds established while constructing their storage. Coverage includes the `[0x19]` hidden-complete-digit case and a tensor-packed witness whose extrema occur after `live_coefficient_len`.

2. **Terminal prewarm route**

   `CommitmentExecutionPlan::inner_ntt_requirement` now derives the route from the execution mode: terminal `InnerOnly` plans produce `CommitmentNttRoute::InnerOnly`, while full and uncompressed plans produce `InnerOuter`. The fused-executor regression test checks that a direct terminal prewarm is dispatched through the split inner registration and is observed as `InnerOnly`.

3. **Owning fold level**

   `CommitmentNttRequirement` now retains `fold_level`, and conversion back to `RoutedNttRequirement` preserves it instead of reconstructing level zero. Route-sensitive test resources verify that level 7 survives both prewarm and retained-cache-policy queries.

4. **Witness-sized copy during freezing**

   `InnerRelationStateMaterial` now holds `Arc<Vec<RingVec<F>>>`, so cloning the carrier is shallow. Freezing takes and consumes the backend inner image, validates the exported rows, and replaces the image with the shared immutable material. Retained-byte coverage verifies that freezing does not increase resident bytes.

   Profiling also showed that this ownership issue was not the source of the measured 25% prover slowdown. The runtime regression came from rescanning every immutable recursive packed witness to rederive bounds. Commit `1eee1c28f` adds the validated packed path and reuses construction-time bounds. The fresh CI benchmark report has all 13 profiles passing, with prover deltas from -4.5% to +1.2%; the recursive W8R2 profile is -0.2%.

5. **Portable resident exports bypassing frozen authority**

   Both `portable_hint` and `into_portable_hint` now go through `frozen_inner_material`. The first access consumes and validates the backend export; transcript binding, relation construction, and later portable exports all share that authority. The fused-state test calls these consumers in sequence and verifies that the inner exporter is consumed once and the portable rows equal the relation rows. Incomplete rows are rejected by the binding-aware material constructor.

6. **Incomplete public material boundary**

   Public `InnerRelationStateMaterial::new` now requires the exact `CommitInnerPlan` and source count and rejects incorrect row count, ring dimension, or coefficient length. The default `InnerRelationState::preflight_inner_relation` materializes and validates generic custom state against the requested plan, so malformed state is rejected by the existing pre-transcript preflight. A custom malformed-state regression verifies that an undersized D=64 row is rejected for a plan requiring 128 coefficients.

The branch also includes the latest `main` at merge commit `c46e48dba`.

quangvdao:
## PR Review findings

Reviewed base `fd249ca64cde7536aad62e23e6eb5496a868046a` through head `c46e48dba2b6033711f8abe7820eb08b1f8085d1`.

Thanks @omibo for the substantial follow-up work here. The original six findings saw meaningful progress: the stored-tail bounds bug is fixed in the canonical view, proof-wide fold/route conversion is repaired, and resident portable exports now share one frozen authority. I found five remaining issues at the current head: one Major and four Moderate.

### 1. Major — public short-norm metadata can diverge from the validated coefficients

`ShortNormRepresentation` maintains two authorities: publicly mutable length/bound fields and a private constructor-validated `packed_view` ([source.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/source.rs#L279-L349)). Materialization validates the public fields, while CPU execution consumes the private view ([compiler.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/source/compiler.rs#L109-L117), [commitment.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/compute/cpu/commitment.rs#L127-L149)).

A safe custom source can construct a valid 512-coefficient representation, then rewrite the public extent fields to advertise 64 coefficients. Compilation accepts it, but the CPU still decodes the private 512-coefficient view and commits a nonzero coefficient at index 64. An exact-head scratch reproduction produced different commitment rows from an advertised all-zero 64-coefficient source.

**Why it matters:** commitment output can depend on coefficients outside the source's advertised domain, violating the PR's primary extension-boundary contract.

**Required correction:** make the metadata private/read-only and make compiler validation and dispatch consume one canonical validated representation. Add the custom-source hidden-tail regression.

### 2. Moderate — direct and runtime cache-policy requests still fabricate fold zero

`CommitmentExecutionPlan::for_recursive` receives and validates `fold_level` but does not retain it ([plan.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/plan.rs#L167-L178)). Consequently, every plan-derived inner and outer requirement is constructed with fold `0` ([resources.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/resources.rs#L141-L194)).

These requirements are used during actual execution, not merely optional direct prewarming ([executor.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/executor.rs#L211-L230)). A level-sensitive backend can therefore make one decision during proof-wide prewarming and the opposite decision immediately before execution.

A scratch regression constructed a real recursive plan at level 1 and expected its requirement to retain level 1; current head returned 0.

**Required correction:** preserve the owning compute-stack level in the execution plan or resource-preparation input, including terminal commitments. Add nonzero recursive and terminal tests comparing proof-wide, direct-prewarm, and runtime identities.

### 3. Moderate — fused-only executors accept an impossible `InnerOnly` prewarm route

When no split registrations exist, `routed_resources` falls back to fused resources even for `InnerOnly` ([resources.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/resources.rs#L418-L449)). The same executor cannot execute that route because `execute_inner` requires a split-inner registration.

A scratch negative test submitted an `InnerOnly` routed requirement to a fused-only executor. Prewarming returned `Ok`, although execution would reject the route. The main proving entry's preflight catches the missing operation, but the public prewarm boundary still contradicts the specification's "same route controls prewarm and execution" rule.

**Required correction:** remove the fused fallback for `InnerOnly`, reject impossible `InnerOnly + Outer` combinations, and add a fused-only negative test.

### 4. Moderate — terminal consumers still create witness-sized copies after freezing

Freezing stores a shared `Arc` in resident state ([state_policy.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/state_policy.rs#L314-L336)). However, `into_rows` deep-clones every row whenever that `Arc` is not unique ([state_policy.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/state_policy.rs#L419-L422)).

Both terminal paths trigger that fallback:

- transcript serialization consumes a cloned carrier merely to inspect one row;
- terminal suffix extraction calls `into_rows` while `prover_state` still retains the cached allocation ([suffix.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/protocol/core/suffix.rs#L272-L283)).

The copied `t_state` then remains live during terminal proving alongside the cached original. The retained-byte test misses this because the temporary transcript copy has already been dropped when accounting runs and it does not exercise terminal suffix ownership.

**Required correction:** serialize through a borrowed row and consume/drop resident state before the final owned-row handoff, or carry the shared allocation through the relation. Add pointer/allocation and terminal peak-retention regressions.

### 5. Moderate — generic inner material is neither exact-plan-bound nor frozen across preflight

`InnerRelationStateMaterial::new` accepts an exact plan but stores only `ring_dimension` and rows. Later validation checks ring dimension, source count, and the product `live_blocks × n_a × D`, not exact plan equality ([state_policy.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/commitment/state_policy.rs#L359-L440)).

Thus these distinct plans alias:

- `D=64, live_blocks=1, n_a=2`
- `D=64, live_blocks=2, n_a=1`

Both accept one 128-coefficient row, although downstream row partitioning treats them differently.

Separately, default generic preflight materializes, validates, and discards the result. After transcript mutation begins, the root relation requests the material again ([prove.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/protocol/core/prove.rs#L102-L146), [ring_relation.rs](https://github.com/LayerZero-Labs/akita/blob/c46e48dba2b6033711f8abe7820eb08b1f8085d1/crates/akita-prover/src/protocol/ring_relation.rs#L442-L453)). A stateful custom implementation can return valid rows during preflight and malformed or different rows at consumption, so rejection occurs only after transcript mutation.

**Required correction:** store and compare the exact consumption-relevant plan and source count, and thread the preflighted material forward as the canonical frozen value. Terminal binding should derive from that same object.

## Earlier-review status

| Earlier finding | Current assessment |
|---|---|
| 1. Encoded tail omitted from extrema | Original bounds/kernel defect fixed; public mutable metadata leaves a distinct hidden-extent path |
| 2. Terminal prewarm route | Fixed for split/hybrid executors; fused-only routed prewarm remains incorrect |
| 3. Fold identity lost | Proof-wide conversion fixed; direct/runtime preparation still rewrites it to zero |
| 4. Persistent witness copy | Backend-image duplication fixed; terminal consumers still create full-row copies |
| 5. Portable exports bypass frozen authority | Fixed |
| 6. Incomplete public material boundary | Not fully fixed: exact-plan aliasing and discard/recompute preflight remain |

`stored_coefficient_len`, route/fold identity, and shared frozen material are earned abstractions. The remaining defects come from parallel authorities around them: mutable metadata versus `packed_view`, proof-wide fold identity versus plan-derived fold zero, and preflight material versus later rematerialization. Repairs should consolidate these authorities rather than add forwarding helpers.

## Validation

- Full 16-file diff reviewed: 627 additions, 169 deletions.
- Historical review at `d44f72b` and the six-point resolution response were checked.
- Relevant source, packed-digit, resource, field-reduction, state-binding, and fusion tests passed.
- Exact-head scratch regressions reproduced the mutable-metadata commitment mismatch, fold-zero rewrite, fused-only routing acceptance, and non-unique-`Arc` row copy.
- The existing current-head benchmark report records 13/13 profiles passing.
- At publication time, completed checks were successful; `Schedule artifact drift` and one benchmark shard rerun were still in progress.

omibo:
Thanks for the careful follow-up. I verified all five findings against `c46e48dba`; they were valid and consistent with the earlier reviews. The fixes are in [`1228e823b`](https://github.com/LayerZero-Labs/akita/commit/1228e823ba11920c84c237971a9e2b3fdea1a867).

### 1. Short-norm metadata and packed view now have one authority

`ShortNormRepresentation`'s validated fields are private and exposed through read-only accessors. Compiler admission and CPU dispatch now consume the same immutable representation, so a custom source cannot construct a larger validated view and then rewrite its advertised byte slice, extents, digit width, or extrema. The packed fast path still reuses construction-time bounds and does not rescan Akita-owned immutable witnesses.

Regression coverage checks the canonical metadata/view relationship and the hidden-tail cases.

### 2. Execution plans retain the owning fold

`CommitmentExecutionPlan` now stores the compute-stack fold that owns the work. Recursive construction checks the expected logical successor while retaining the owner, and terminal construction receives its fold explicitly. Inner and outer NTT requirements derive their fold from the plan rather than fabricating fold zero.

Tests exercise nonzero recursive and terminal folds and compare the identities seen by proof-wide planning, direct preparation, and runtime preparation.

### 3. Resource routing rejects operations the executor cannot run

`InnerOnly` now requires a split-inner registration; it no longer falls back to fused resources. `InnerOnly + Outer` is rejected before resource-policy callbacks. Fused-only full/uncompressed work remains supported, as does a fused-plus-split-inner configuration for terminal work.

Negative tests cover fused-only `InnerOnly` and the impossible `InnerOnly + Outer` combination.

### 4. Terminal and nonterminal consumers avoid witness-sized row copies

Terminal transcript serialization borrows the frozen row instead of consuming a cloned carrier. Before the final owned terminal handoff, the prover drops the resident owner and then extracts the rows, allowing the unique `Arc` allocation to move without a deep clone. The nonterminal ring-switch path now borrows rows during decomposition as well.

The copying fallback remains available for an explicit portable export while shared leases are still alive.

### 5. Generic material is exact-plan-bound and prepared once

`InnerRelationStateMaterial` now retains the exact `CommitInnerPlan` and source count and validates both on use, so same-product plans such as `(live_blocks=1, n_a=2)` and `(live_blocks=2, n_a=1)` no longer alias.

The proof prepares and validates generic inner material once before transcript mutation, then carries that exact object into transcript binding and relation construction. It is no longer discarded and rematerialized after the transcript has advanced. Tests cover same-product/different-plan rejection, stateful material providers, malformed initial material, and single materialization.

### Related audit fixes

- Compression material now validates relation mode, exact plan, quotient count, and quotient geometry before use.
- A failed resident freeze is represented explicitly and returns a stable error on later access instead of becoming an ambiguous empty state.
- Recursive commitment output carries the already prepared inner/compression material forward, keeping preparation and consumption on the same authority.

Validation completed successfully:

- `cargo test -p akita-prover --lib` — 435 passed, 3 ignored
- `cargo test -p akita-pcs --test akita_small_field_e2e` — 9 passed, 1 ignored
- focused commitment-contract and external-backend tests
- workspace all-target compilation
- all three CI Clippy feature configurations
- formatting, repository scripts, dependency checks, Machete, typos, and documentation guardrails

quangvdao:
## PR re-review findings

Re-reviewed base `fd249ca64cde7536aad62e23e6eb5496a868046a` through head `1228e823ba11920c84c237971a9e2b3fdea1a867`.

Thanks @omibo for the thorough follow-up. The five findings from my previous review are addressed at this head: short-norm metadata now has one immutable authority, execution plans preserve the owning fold, fused-only executors reject `InnerOnly`, terminal resident rows avoid the deep-copy fallback, and generic relation material is exact-plan-bound and prepared before transcript mutation.

The incremental re-review found two Moderate issues and one Low issue.

### 1. Moderate — an inner-only executor accepts the symmetric impossible `InnerOuter` prewarm route

The new routing check correctly rejects `InnerOnly` on a fused-only executor. The symmetric case remains: for `(InnerOuter, Inner)`, `routed_resources` returns the split-inner resources without verifying that the executor also has an outer registration ([resources.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/commitment/resources.rs#L424-L456)). The builder legitimately permits an inner-only executor ([builder.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/commitment/builder.rs#L129-L164)), while actual `InnerOuter` execution requires the missing outer registration ([executor.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/commitment/executor.rs#L282-L315)).

A focused scratch regression built a legal inner-only executor, derived the inner requirement from a full root plan, and submitted it as `InnerOuter + Inner`; prewarming returned `Ok`, although that executor cannot execute the plan. A complete request can therefore mutate/cache the inner stage before failing on its later outer requirement.

**Required correction:** validate the complete prepared route before returning stage resources: `InnerOuter` must require either a split inner+outer route or a fused route, while `InnerOnly` must require the terminal split-inner operation. Add the symmetric inner-only/`InnerOuter` negative regression.

### 2. Moderate — eager resident compression preparation retains two full compression images across the following fold

`commit_w` now keeps the resident `prover_state` and also exports prepared compression material ([commit.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/protocol/ring_switch/commit.rs#L80-L105)). For the CPU resident path, borrowed export deep-clones both the compression-chain witness and every quotient row ([compression_stage.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/compute/cpu/compression_stage.rs#L59-L80)). Both the original backend state and copied material are then carried into the next `SuffixProverState` ([fold/mod.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/protocol/core/fold/mod.rs#L673-L715)).

The previous implementation briefly held both copies during relation construction; eager preparation now lengthens that duplication so it overlaps the next fold's preparation and proving work. It also leaves `ResidentCommitmentState::retained_bytes()` accounting only the original backend state, not the prepared duplicate.

**Required correction:** freeze or consume compression into one canonical shared material, analogous to inner state, or drop the original compression state once the prepared material has been validated and carried forward. Add a retained/peak-allocation regression covering the recursive resident path.

### 3. Low — portable compression preflight still reconstructs and discards complete material

`batched_prove` calls compression preflight and then prepares the same material immediately afterward ([prove.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/protocol/core/prove.rs#L99-L123)). For `AkitaCommitmentHint`, preflight delegates to `outer_compression_material(...).map(drop)` ([state_policy.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-prover/src/commitment/state_policy.rs#L592-L621)); rebuilding that material clones packed stages and quotient rows ([hints.rs](https://github.com/LayerZero-Labs/akita/blob/1228e823ba11920c84c237971a9e2b3fdea1a867/crates/akita-types/src/proof/hints.rs#L137-L191)). Preparation then repeats the work.

This is no longer a TOCTOU problem because the hint is immutable, but it contradicts the new cheap-preflight contract and adds avoidable copying.

**Required correction:** make portable preflight validate shapes and mode without reconstructing owned material; construct the frozen material only once in `prepare_commitment_relation_material`.

## Validation

- `cargo test -p akita-prover --lib`: 435 passed, 3 ignored.
- `cargo test -p akita-pcs --test commitment_contract`: passed.
- Focused commitment tests and the new fold/route regressions passed.
- The symmetric inner-only/`InnerOuter` scratch regression failed as described because prewarming returned `Ok`.
- At publication time, no upstream check was failing; remaining in-progress checks are reported separately by GitHub.

omibo:
Thanks for the re-review. I verified all three findings; they were valid and consistent with the earlier routing, frozen-state, and cheap-preflight requirements. They are addressed in [`bb2d040d8`](https://github.com/LayerZero-Labs/akita/commit/bb2d040d8).

### 1. Symmetric `InnerOuter` route validation

Resource selection now validates the complete executable route before returning resources or invoking a cache callback:

- `InnerOnly` requires a split-inner operation and rejects an outer-stage request.
- `InnerOuter` requires either a fused operation or both split-inner and split-outer operations.

The new inner-only-executor regression submits an `InnerOuter + Inner` requirement, verifies rejection, and verifies that the resource callback count remains zero.

### 2. Resident compression now freezes into one shared allocation

Resident compression state now has the same explicit `Image -> Material` ownership transition as resident inner state. Its first material request consumes the opaque backend state, validates the exact compression plan, relation mode, quotient count, and quotient geometry, and stores a shared immutable material carrier. Later requests clone only the carrier.

`commit_w` therefore carries a reference to the same compression allocation retained by `prover_state`; it no longer keeps a copied host image beside the backend image. Ring-relation construction also retains that shared carrier instead of extracting owned witness and quotient vectors.

The regression verifies that repeated relation-material requests share the allocation, the consuming exporter runs once, and retained bytes do not increase across freezing.

### 3. Portable compression preflight is borrowed

`AkitaCommitmentHint` preflight now checks stage count, packed-byte lengths and padding, relation mode, quotient count, ring dimensions, and coefficient lengths directly through borrowed slices. It does not reconstruct `CompressionChainWitness` or clone quotient rows. Owned material is built once later at `prepare_commitment_relation_material`.

The existing public reconstruction methods remain available at the actual export/materialization boundary.

Validation completed successfully:

- `cargo test -p akita-prover --lib` — 435 passed, 3 ignored
- `cargo test -p akita-types --lib` — 608 passed, 1 ignored
- `cargo test -p akita-pcs --test akita_small_field_e2e` — 9 passed, 1 ignored
- `cargo test -p akita-pcs --test commitment_contract` — passed
- workspace-wide all-target compilation
- all three CI Clippy feature configurations
- formatting, line limits, dependency hygiene, shared-field identity, typos, and documentation guardrails

quangvdao (approved):
Final re-review at bb2d040d8: no actionable findings. All findings from the earlier review passes are addressed, and the exact-head validation passed. Thanks Omid for the thorough fixes and careful follow-through.

RadNi (approved):

PR #27, opened 2026-09-12
fix(estimator): use q as the Euclidean SIS triviality bound

quangvdao:
## Summary

This PR corrects Akita's Euclidean-SIS triviality boundary from `(q - 1) / 2` to `q`.

For standard Euclidean SIS, the unconditional modular solution is `q * e_i`, whose L2 norm is exactly `q`. Crossing the centered-representative boundary `(q - 1) / 2` does not by itself yield a trivial SIS solution. The upstream lattice estimator made the same correction in [malb/lattice-estimator#218](https://github.com/malb/lattice-estimator/pull/218), following the analysis in [malb/lattice-estimator#162](https://github.com/malb/lattice-estimator/issues/162).

### Current diff metadata

- Base: `c2a70dc68c489d28dfaba7bf1cc897cd9823e107` (`main` at the final branch update)
- Head: `6c1137d8933f81c36a84bb66579b2d2d3f15ce4e`
- Commits unique to the branch: 3, including the merge from current `main`
- Files changed: 2
- Diff: 160 insertions, 10 deletions

## Motivation

The previous Euclidean estimator rejected every collision bound at or above `(q - 1) / 2` before running the lattice-cost model. That cutoff confuses the centered modular-representative interval with the norm of a trivial Euclidean-SIS witness and can falsely reject secure parameter candidates whose bounds lie between `q / 2` and `q`.

## Changes

- Use `length_bound >= q` as the Euclidean triviality predicate.
- Compare each `Bound` representation without log-space boundary rounding:
  - integer bounds compare directly with `q`;
  - rational bounds compare the numerator with `q * denominator`;
  - square-root bounds compare the radicand with `q^2`;
  - finite positive floating bounds compare their integer floor with `q`.
- Update the rejection message to name the Euclidean `q` boundary.
- Exercise both public estimator entry points with a finite bound between `q / 2` and `q`, and reject public inputs at and above `q`.
- Cover q32 floating values immediately below, at, and above the cutoff; a modulus above exact `f64` integer precision; and exact q64/q128 integer and square-root boundaries.
- Correct stale public API documentation that described the Euclidean estimator as unimplemented.

The infinity-norm estimator is intentionally unchanged; its existing `(q - 1) / 2` policy is a separate norm-specific path.

## Generated artifacts

Generated L2 tables and schedule artifacts are intentionally unchanged. Existing table entries remain conservative, and all current q32 schedules lie below the affected bound range. Regenerating the complete offline L2 table could add newly supportable large-bound rows, but is not required to correct the estimator predicate or preserve the security of existing schedules.

## Security impact

This removes false negatives from Euclidean parameter selection. It does not weaken the lattice cost calculation or accept the unconditional `q * e_i` solution: bounds at or above `q` continue to be rejected before estimation.

## Validation

Validated locally at `6c1137d8933f81c36a84bb66579b2d2d3f15ce4e`:

- `cargo test -p akita-sis-estimator --lib --tests` — 105 tests passed;
- estimator Clippy across all targets and features with warnings denied;
- repository Rust formatting;
- Rust file-line self-tests and repository line-cap check;
- diff whitespace check.

Fresh repository CI at the current head completed with 47 checks passed, one expected Book deployment skip, and no failures.

## Reviewer map

- `crates/akita-sis-estimator/src/euclidean.rs`: exact boundary predicate, diagnostic, public-path tests, and representation/precision boundary tests.
- `crates/akita-sis-estimator/src/lib.rs`: corrected public Euclidean API error documentation.

### DIFF
--- a/crates/akita-sis-estimator/src/euclidean.rs
+++ b/crates/akita-sis-estimator/src/euclidean.rs
@@ -1,14 +1,14 @@
 //! Euclidean-norm SIS lattice cost.
 
 use num_bigint::BigUint;
-use num_traits::One;
+use num_traits::FromPrimitive;
 
 use crate::{
     config::{EstimateConfig, ReductionCostModel},
     cost::{CostValue, EstimateTag, LatticeCost},
     error::{EstimatorError, Result},
     math::log2_biguint,
-    params::SisParameters,
+    params::{Bound, SisParameters},
     reduction::{adps16_log2_cost, beta as beta_from_delta, delta, validate_euclidean_reduction},
 };
 
@@ -18,7 +18,7 @@ pub fn cost_euclidean(params: &SisParameters, config: &EstimateConfig) -> Result
     if length_bound_trivially_easy(params) {
         return Err(EstimatorError::InvalidParameter {
             field: "length_bound",
-            reason: "SIS trivially easy: length_bound must be below (q - 1) / 2".to_string(),
+            reason: "SIS trivially easy: Euclidean length_bound must be below q".to_string(),
         });
     }
 
@@ -99,8 +99,18 @@ fn opt_sis_dimension(params: &SisParameters, m: u64, log_q: f64) -> Result<u64> 
 }
 
 fn length_bound_trivially_easy(params: &SisParameters) -> bool {
-    let half_q_log2 = log2_biguint(&(params.q.clone() - BigUint::one())) - 1.0;
-    params.length_bound.log2() >= half_q_log2
+    // Standard Euclidean SIS has the unconditional solution q * e_i, whose
+    // length is q.  The centered-representative boundary (q - 1) / 2 is not a
+    // triviality threshold for this norm.
+    match &params.length_bound {
+        Bound::Integer(value) => value >= &params.q,
+        Bound::Float(value) => BigUint::from_f64(*value).is_some_and(|value| value >= params.q),
+        Bound::Rational {
+            numerator,
+            denominator,
+        } => numerator >= &(&params.q * denominator),
+        Bound::SqrtInteger(radicand) => radicand >= &(&params.q * &params.q),
+    }
 }
 
 fn length_bound_exceeds_euclidean_lower_bound(params: &SisParameters, d: u64, log_q: f64) -> bool {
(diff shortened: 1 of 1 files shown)


crates/akita-sis-estimator/src/euclidean.rs
```diff
@@ -195,4 +205,43 @@ mod tests {
             CostValue::finite_log2(adps16_log2_cost(beta, Adps16Mode::Quantum))
         );
     }
+
+    #[test]
+    fn euclidean_triviality_boundary_is_q() {
```
sumchecker, line 210: Small test nit, non-blocking: it'd be good to exercise the public `estimate` / `cost_euclidean` entry points here too. These assertions only call the helper, so they'd still pass if the actual guard disappeared or estimation failed later for a bound between `q/2` and `q`.

A handy fixture is `n=1024, q=12289, m=27824, B=8382.44`: it should give a finite cost (`beta=953`, `d=2134` with ADPS16 quantum). Then check that bounds at and above `q` return a `length_bound` error. I checked these cases on this commit and they pass; it'd just be useful to keep that coverage in the suite.

crates/akita-sis-estimator/src/euclidean.rs
```diff
@@ -195,4 +205,43 @@ mod tests {
...
+        assert!(!length_bound_trivially_easy(&params(Bound::SqrtInteger(
+            &q * &q - BigUint::from(1u8)
+        ))));
+        assert!(length_bound_trivially_easy(&params(Bound::SqrtInteger(
+            &q * &q
+        ))));
+
+        assert!(!length_bound_trivially_easy(&params(Bound::Float(
+            3_000_000_000.0
+        ))));
```
sumchecker, line 242: Another non-blocking test nit: could we cover floats right next to the cutoff, plus a modulus that doesn't fit exactly in `f64`? `3_000_000_000.0` is quite far below q32, and q32 itself is exactly representable.

I'd add q32's `q_f64.next_down()`, `q_f64`, and `q_f64.next_up()`, then `q = 2^53 + 1` with `B = 2^53` (below q) and `B = 2^53 + 2` (reject). That would catch someone accidentally switching to a rounded floating-point modulus later. The current code passes these cases. A couple of boundary cases for q64/q128 would be useful too.

sumchecker (dismissed):
The cutoff correction looks good. Existing generated tables, schedules, and verifier bounds are unchanged, and the estimator checks and CI on this commit pass. Left two small, non-blocking test suggestions.

crates/akita-sis-estimator/src/euclidean.rs
quangvdao: Addressed in 6c1137d8933f81c36a84bb66579b2d2d3f15ce4e: the regression now exercises both public entry points with the suggested finite fixture (beta 953, d 2134) and verifies length_bound rejection at and above q.

crates/akita-sis-estimator/src/euclidean.rs
quangvdao: Addressed in 6c1137d8933f81c36a84bb66579b2d2d3f15ce4e: added q32 next-down/exact/next-up float cases, the q = 2^53 + 1 precision boundary, and exact integer/square-root boundaries for q64 and q128.

quangvdao:
@sumchecker I updated the branch to current main and addressed both non-blocking review suggestions in 6c1137d8933f81c36a84bb66579b2d2d3f15ce4e. All fresh checks are green except the still-running schedule/security gate. A fresh last-push approval is required by the repository rules when you have a moment.

omibo (approved):

PR #29, opened 2026-09-14
Remove stale sumcheck protocol paths

omibo:
Removes stale sumcheck APIs while preserving the active prover, verifier, and batching behavior. Internal callers now use direct protocol functions with the same proof-shape, degree, transcript, challenge, and terminal-claim checks.

| ID | Decision | Surface |
|----|----------|---------|
| E1 | Removed | Standard prover/verifier extension traits and `check_sumcheck_output_claim`; direct single-instance functions replace internal method calls. |
| E2 | Removed | Equality-factored prover/verifier extension traits; direct prove/verify functions preserve the normalized claim recurrence. |
| E3 | Removed | Equality-factored verifier and mutable round-state traits. |
| E4 | Retained | Root `prove_batched_sumcheck` and `verify_batched_sumcheck` for external consumers. |
| E5 | Retained | Split batched verification, result metadata, expected-output computation, and output-claim checking for external consumers. |
| E6 | Removed | Inherent `SumcheckProof::verify`; direct verification and retained batching share canonical round-message validation. |
| E7 | Retained | `GruenSplitEq::try_gruen_poly_from_coeffs_except_linear` for external consumers. |

Also removes the unused range-check evaluator, small-table MLE evaluator, virtual-only stage-2 constructor, and duplicate module-qualified sumcheck exports. The equality-factored claim recurrence remains private to the canonical prover/verifier implementation.

Validation: focused sumcheck and prover tests, all three release Clippy configurations, formatting, dependency checks, line limits, script tests, schedule regeneration, and repository preflight checks.


quangvdao:
## PR Review findings

Reviewed the complete `c2a70dc68c489d28dfaba7bf1cc897cd9823e107...a55a41ceb108e37feb6fcc0168175e6120570469` surface (34 files, +741/-1,895).

### 1. Moderate — the live Book still documents interfaces deleted by this PR

The Book says that public `advance_eq_factored_claim` lives in `src/drivers/eq_factored.rs` and that both the prover and verifier use the factored interfaces there ([lines 114-117](https://github.com/LayerZero-Labs/akita/blob/a55a41ceb108e37feb6fcc0168175e6120570469/book/src/foundations/eq-factored-sumcheck.md#L114-L117)). Its implementation map likewise points reviewers to factored prover/verifier interfaces and the deleted driver ([lines 167-175](https://github.com/LayerZero-Labs/akita/blob/a55a41ceb108e37feb6fcc0168175e6120570469/book/src/foundations/eq-factored-sumcheck.md#L167-L175)).

At this head, the recurrence is a private helper shared by the direct functions in [`single.rs`](https://github.com/LayerZero-Labs/akita/blob/a55a41ceb108e37feb6fcc0168175e6120570469/crates/akita-sumcheck/src/single.rs#L175-L275), and the factored verifier trait has been removed.

**Why it matters:** this leaves the canonical narrative and reviewer map pointing at a deleted file and describing an ownership model that the PR intentionally removes.

**Required correction:** update the Book to identify `single.rs`, the retained factored prover trait, and the direct `verify_eq_factored_sumcheck` contract.

### 2. Moderate — security-relevant eq-factored regression cases are deleted without equivalent replacements

The deleted driver suite independently covered the historical old-wire forgery at `tau = 0`, exact headerless transmission of every nonconstant coefficient, valid degree-zero factored messages, and Boolean/non-Boolean equality coordinates ([base lines 109-216](https://github.com/LayerZero-Labs/akita/blob/c2a70dc68c489d28dfaba7bf1cc897cd9823e107/crates/akita-sumcheck/tests/drivers.rs#L109-L216)).

The replacement retains the roundtrip/terminal check and the important later-round tamper test after an equality factor vanishes ([head lines 209-248](https://github.com/LayerZero-Labs/akita/blob/a55a41ceb108e37feb6fcc0168175e6120570469/crates/akita-sumcheck/src/single/tests.rs#L209-L248)), but it does not retain the exact `tau = 0` forgery or degree-zero/wire invariants.

**Why it matters:** the current recurrence remains correct, but a prover/verifier roundtrip is not an independent oracle against reintroducing the old shared wire bug.

**Required correction:** port at least the `tau = 0` forgery test and the exact serialization/degree-zero cases to the new direct API.

### 3. Low — malformed full verification now mutates the transcript before rejection

The old full verifier validated every round message before absorbing the input claim ([base lines 128-148](https://github.com/LayerZero-Labs/akita/blob/c2a70dc68c489d28dfaba7bf1cc897cd9823e107/crates/akita-sumcheck/src/drivers/standard.rs#L128-L148)); the deleted test required invalid raw and full verification to sample nothing and leave the transcript unchanged ([base lines 374-439](https://github.com/LayerZero-Labs/akita/blob/c2a70dc68c489d28dfaba7bf1cc897cd9823e107/crates/akita-sumcheck/tests/drivers.rs#L374-L439)).

The new `verify_sumcheck` absorbs the claim and only then calls the validating replay function ([head lines 146-161](https://github.com/LayerZero-Labs/akita/blob/a55a41ceb108e37feb6fcc0168175e6120570469/crates/akita-sumcheck/src/single.rs#L146-L161)). I reproduced that an invalid-size proof returns the expected error but leaves a different transcript state from a fresh transcript.

**Why it matters:** current PCS callers abort immediately, so this is not an invalid-proof acceptance or soundness issue. It is nevertheless an observable failure-path change that the old tests explicitly guarded.

**Required correction:** restore validation before claim absorption without duplicating the protocol authority (for example, share a private validated replay primitive), or explicitly accept and test the new error-side-effect contract.

## Overall assessment

I found no valid-proof correctness, verifier no-panic, or transcript/challenge-order defect. The main refactor is sound: the blanket extension traits and wrapper verifier objects are replaced by direct canonical functions while proof encodings, degree checks, round recurrences, terminal claims, EOR/Stage 3 split replay, and front-loaded batching remain intact.

One scope note: the source break is broader than E1-E7. It also removes the public `range_check_eval`, `multilinear_eval_small`, nested module import paths, and the only public `RelationRangeImageProver` constructor (`new_virtual_only`). There are no in-tree callers and Akita disclaims backward compatibility, so this is not a correctness blocker, but it is worth enumerating in the PR body for external consumers.

Focused exact-head validation passed for `akita-sumcheck`, `akita-verifier --lib`, and `akita-prover --lib`; hosted checks at this head currently report 47 successes and one intentional skip.

Thanks @omibo — the consolidation toward one direct sum-check driver per concept is a meaningful simplification. The remaining asks are narrowly about keeping the Book and security regression coverage aligned with it.

RadNi (approved):

PR #31, opened 2026-09-16
refactor(prover): internalize batch fold fallback

RadNi:
## Summary

- remove `BatchDecomposeFoldOutcome` and make batched opening kernels return the aggregate witness directly
- move per-polynomial fallback into dense and recursive backend implementations while retaining the one-hot fused path
- centralize checked witness aggregation and uniform batch challenge validation
- remove the protocol-layer fallback branch and its unnecessary single-fold backend constraint

## Validation

- `cargo fmt --all --check`
- `taplo fmt --check`
- all three Clippy configurations from `.github/workflows/ci.yml`
- crate dependency hygiene checks from the Clippy job
- release Rustdoc with warnings denied
- `./scripts/check-doc-guardrails.sh`
- `python3 scripts/tests/test_doc_blast_radius.py`
- `rtk cargo nextest run -p akita-prover --lib` (435 passed, 3 skipped)


quangvdao:
## PR Review findings

Reviewed `f6c3643bddd5ecff9270fe843c8e58d0514f78d3...de30bf4b744cf33237967a587eee8f4310686fc3` (11 files, +212/−145).

### 1. Moderate — uniform batch validation accepts mixed layouts and produces backend-dependent witnesses

[`DecomposeFoldBatchPlan::challenges_per_poly`](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/compute/operation_plans.rs#L280-L300) checks only that the polynomial and challenge collections are nonempty and that the challenge count is divisible by the polynomial count. It does not establish that every source has the same live-block extent.

The built-in backends consequently assign the same malformed batch different meanings:

- [dense splits the challenges evenly](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/backend/dense/kernels.rs#L84-L103);
- [one-hot calculates that split but does not use it on the fused path](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/backend/onehot/ops.rs#L216-L243), instead [consuming challenges according to each source's actual block count](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/backend/onehot/decompose_fold.rs#L219-L246).

**Reproduction:** I constructed equivalent dense and one-hot batches whose two polynomials have two and four live blocks, then supplied six challenges. Both batch kernels returned `Ok`; dense partitioned the challenges 3/3 while one-hot used 2/4, and the resulting witnesses differed.

There is a related unchecked pairing at the protocol boundary: [`build_point_decompose_fold_witness`](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/protocol/ring_relation.rs#L247-L258) does not require `point_polys.len() == point_indices.len()` before lowering `Challenges`, which carries the authoritative claim layout, to a raw slice. For example, one polynomial plus two selected claims is silently reinterpreted as both claims' challenges belonging to that polynomial.

**Why it matters:** normal production preparation currently creates aligned homogeneous groups, so this is not an active verifier-soundness break. However, the new public kernel contract claims uniform validation while accepting malformed input and returning backend-dependent results.

**Required correction:** retain an authoritative per-polynomial challenge count in the plan, validate the checked total and source extents, reject polynomial/index count disagreement, and make the one-hot fused path consume exactly the validated partitions. Add mixed-shape and count-mismatch rejection tests for dense, one-hot, setup-prefix, and suffix-witness batches.

### 2. Moderate — downstream backends cannot reuse the canonical fallback aggregation

The new [`OpeningBatchKernel` contract](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/compute/kernels.rs#L97-L112) requires a backend without a fused path to fold its sources individually and return the final aggregate witness.

The PR correctly creates one canonical implementation with ring-dimension, row-count, and checked-`i32` accumulation validation ([implementation](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/compute/kernels.rs#L11-L56)), but [exports it only as `pub(crate)`](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/crates/akita-prover/src/compute/mod.rs#L48-L53).

A downstream source-typed backend that previously returned `FallbackPerPoly` must therefore duplicate this aggregation logic or invent another version. The source break itself is permitted by Akita's no-compatibility policy, but requiring parallel implementations conflicts with the repository's [one-canonical-function rule](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/AGENTS.md#L83-L85).

**Required correction:** expose the checked aggregation primitive through the public compute API, or redesign the trait so Akita continues to own generic fallback aggregation while the backend owns source-specific scalar folding. Add a downstream-like custom batch-kernel contract test.

### 3. Low — the deleted public outcome type is missing from the required documentation guards

The PR removes the public `BatchDecomposeFoldOutcome` type. That breaking change is allowed, but repository policy requires removed public API names to be added to both deleted-symbol lists ([policy](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/docs/documentation.md#L104-L119)). The symbol is absent from the [documentation guard list](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/scripts/check-doc-dead-symbols.sh#L100-L139) and [specification guard list](https://github.com/LayerZero-Labs/akita/blob/de30bf4b744cf33237967a587eee8f4310686fc3/scripts/check-spec-references.sh#L108-L130).

**Failure mode:** live documentation can retain or reintroduce the removed API without the guardrails noticing it.

**Required correction:** add `BatchDecomposeFoldOutcome` to both deleted-API pattern lists. Archived specifications should remain historical; a short live architecture note should record that batch kernels now own fallback aggregation.

## Specification alignment

| Requirement | Assessment |
|---|---|
| Remove `BatchDecomposeFoldOutcome`; return the final witness from the batch kernel | Met |
| Preserve the one-hot fused path | Met |
| Internalize dense, one-hot, setup-prefix, and suffix-witness fallbacks | Met for well-formed production batches |
| Centralize checked witness aggregation | Met in-tree; unavailable to downstream implementors (finding 2) |
| Uniform challenge validation | **Not fully met:** mixed source extents and polynomial/index count mismatches are accepted (finding 1) |
| Remove the protocol fallback branch and its local scalar-fold bound | Met |
| Public API/documentation cutover | Type removal is permitted; mandatory deleted-symbol guards are missing (finding 3) |

The ownership direction is sound: the protocol no longer negotiates backend fallback behavior, and the shared aggregation function owns real checked arithmetic rather than acting as a forwarding wrapper. The remaining abstraction issues are the incomplete batch-layout contract and the canonical helper's crate-private visibility; I found no additional wrapper or indirection concerns.

## Validation

- Exact-head local prover suite: 435 passed, 3 ignored.
- Focused scratch reproduction confirmed finding 1.
- At the final refresh, all GitHub test, Clippy, fuzz, portability, transcript, security, documentation, schedule-artifact, and benchmark checks had completed successfully; the Book deployment job was intentionally skipped.
- For well-formed production groups, the dense/setup-prefix/suffix scalar call order, prepared-setup propagation, aggregation order, overflow checks, and one-hot fused arithmetic are preserved.

crates/akita-prover/src/backend/onehot/decompose_fold.rs
```diff
@@ -210,33 +210,34 @@ impl<F: Field, I: OneHotIndex> OneHotPoly<F, I> {
...
-            if challenge_start == challenges.len() {
-                break;
-            }
             let (ring_elems, num_blocks) = poly.view_layout(D, num_positions_per_block).ok()?;
-            let active_blocks = num_blocks.min(challenges.len() - challenge_start);
-            if active_blocks == 0 {
-                continue;
+            if num_blocks != challenges_per_poly {
+                return None;
             }
```
sumchecker, line 233: Is it expected that passing the wrong number of challenges returns `Ok` with an all-zero witness?

For example, a two-block one-hot polynomial expects two challenges, but passing one or three now hits this `None`, which the scalar caller turns into a zero witness. If that's not intended, could we return `AkitaError` for the count mismatch instead?

crates/akita-prover/src/backend/onehot/ops.rs
```diff
@@ -212,23 +212,42 @@ where
...
+                num_positions_per_block,
+                num_digits,
+                log_basis,
+            ) {
+                Some(witness) => Ok(witness),
+                None => aggregate_decompose_fold_witnesses::<F, D>(
+                    source
+                        .polys
+                        .iter()
+                        .zip(window.chunks_exact(challenges_per_poly))
```
sumchecker, line 239: Do we still need this fallback?

Once the batch and one-hot geometry checks pass, the fused helper should always return a witness. Folding each polynomial separately doesn’t give us another way to handle a failure: the scalar method calls the same helper again, just with one polynomial.

If we make the fused helper return `Result` and pass the error back, then we can drop this extra loop.

crates/akita-prover/src/backend/dense/tests.rs
```diff
@@ -139,3 +143,79 @@ fn dense_field_constructor_rejects_arity_that_truncates_to_a_valid_shift() {
...
+        )
+    };
+
+    let chunks = run(2).unwrap();
+    let global = run(1).unwrap().pop().unwrap();
+    let recombined =
+        aggregate_decompose_fold_witnesses::<F, D>(chunks.iter().cloned().map(Ok)).unwrap();
+
+    assert_eq!(chunks.len(), 2);
+    assert_eq!(recombined, global);
```
sumchecker, line 220: Can we check each chunk’s expected value too? Swapping the chunks or returning `[global, zero]` would still pass this test, since their sum is unchanged. The protocol uses each chunk separately, so the test should catch both cases.

crates/akita-prover/src/backend/onehot/ops.rs
mtbadakhshan: Reviewing this together with #36, this contract problem survives the backend extraction in `akita-cpu-backend/src/sources/onehot/ops.rs`.

The scalar fallback calls the same fused helper again, so it does not provide a different execution strategy. The exact-count check also returns `None` for a two-block source given one or three challenges, and the scalar caller converts that into an all-zero witness. Validated batch callers reject malformed counts earlier, but the lower-level contract still hides the failure.

Please return `Result` from the fused helper, propagate invalid geometry/count errors, and delete the one-hot per-polynomial fallback. Define any intentional empty-source behavior explicitly. Add short/excess challenge-count tests at the scalar boundary and assert each chunk's expected value, not only their sum.

sumchecker (approved):

crates/akita-prover/src/backend/recursive/witness.rs
```diff
@@ -622,11 +622,45 @@ where
...
         &self,
-        _prepared: Option<&Self::PreparedSetup>,
-        _source: SuffixWitnessBatchView<'_, F, D>,
-        _plan: DecomposeFoldBatchPlan<'_>,
-    ) -> Result<BatchDecomposeFoldOutcome<F, D>, AkitaError> {
-        Ok(BatchDecomposeFoldOutcome::FallbackPerPoly)
+        prepared: Option<&Self::PreparedSetup>,
+        source: SuffixWitnessBatchView<'_, F, D>,
+        plan: DecomposeFoldBatchPlan<'_>,
+    ) -> Result<Vec<DecomposeFoldWitness<F>>, AkitaError> {
```
mtbadakhshan: **The decomposition concern remains in this lower PR.** Re-reviewed at `22353d52`: this file still grows from 944 lines on the current base to 1,018 lines here. It combines recursive witness representation, source views, backend kernel implementations, and tests.

The ordinary-module decomposition in #36 addresses the final stacked state, but #31 still introduces the oversized module on its own. Please bring the relevant decomposition into this PR so the batch-kernel changes have a focused module boundary before #31 merges. This is the remaining lower-PR portion of the earlier [stack decomposition finding](https://github.com/LayerZero-Labs/akita/pull/36#discussion_r4066298219); the `include!` issue in #36 is fixed.

PR #36, opened 2026-09-20
refactor(prover): isolate an owning CPU backend behind opaque kernels

RadNi:
Protocol orchestration now uses opaque backend contracts while one reusable `CpuBackend` owns sources, retained commitment material, witness storage, arithmetic, setup resources, and caches. Commitments retain their exact sources, and independent proof sessions allow concurrent proofs to share immutable commitments without sharing mutable round state.

- Move CPU commitment/opening, EOR, recursive witness construction, and Stages 1–3 into `akita-cpu-backend`; enforce that `akita-prover` has no production dependency on it.
- Validate ownership, setup, commitment/statement agreement, operation lineage, and session progression. Keep private witness contents behind handles and release only scheduled protocol messages.
- Migrate PCS/application callers, setup-prefix persistence, diagnostics, tests, benchmarks, documentation, and CI guards. Preserve transcript ordering and deterministic proof behavior.
- Cover the generic entrypoint with an external backend fixture and exercise shared commitments, independent sessions, cleanup, and invalid handle combinations.

Validation completed during implementation: both CI test shards (1,844 passing tests), all three required Clippy configurations with warnings denied, repository preflight, transcript and ownership suites, diagnostics, Jolt integration, fuzz smoke runs, and the Wasm check. Five passing tests reported output-handle warnings in parallel runs; all five passed without warnings when rerun sequentially. Linux/x86 validation remains for CI because local target installation was declined.

Replaces #33 after renaming its source branch to `codex/owning-cpu-backend`; retains the same base branch.


quangvdao:
## PR Review findings

Reviewed **`ea74e9544b69424fb32d702024edf7d2c12dd589`** against base **`4215b5c2bec75fb2810905a6e6c57e4e79d530e6`** (`codex/remove-batch-decompose-fallback`, not main). This includes the two latest performance commits, `b8d7997` and `ea74e95`, and the full 411-file change surface. Three actionable findings follow. None is a demonstrated verifier-soundness vulnerability.

### 1. Moderate — sign-insensitive extrema checks reject a valid negative endpoint

In [fold admission](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/fold.rs#L386-L400), `abs(min)` is checked against the negative bound while `abs(max)` is checked against the positive bound. When every coefficient is negative, the maximum is still a negative coefficient and must use the negative bound.

For balanced base-256 digits, the legal interval is `[-128, 127]`. A response `[-128; 64]` has `min = max = -128`: every coefficient is representable, but the new predicate rejects it because `128 > 127`. The [actual extrema calculation](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/decompose_fold.rs#L119-L136) does not clamp the maximum to zero. The [base implementation](https://github.com/LayerZero-Labs/akita/blob/4215b5c2bec75fb2810905a6e6c57e4e79d530e6/crates/akita-prover/src/protocol/fold_grind.rs#L61-L82) used a sign-aware coefficient predicate.

**Impact:** extra rejection/retries and potentially a different accepted nonce, transcript and proof bytes. This violates the stated acceptance/nonce-preservation contract, even though the corner case may be rare in ordinary large random responses. Both sparse and per-chunk admission call this helper. It is a completeness/parity defect, not permissive acceptance of invalid proofs.

**Validation:** compiled and executed a standalone Rust probe containing the current admission function verbatim, with minimal field-free getter fixtures. For constant vectors at `-129, -128, -127, 0, 127, 128`, only `-128` disagreed with the independent interval predicate: expected acceptance, actual rejection. This was a predicate-level reproduction, not an end-to-end PCS proof fixture.

**Repair:** compare signs as well as magnitudes:

```rust
(min < 0 && u128::from(min.unsigned_abs()) > negative_bound)
    || (max > 0 && u128::from(max.unsigned_abs()) > positive_bound)
```

Alternatively reuse a single sign-aware coefficient predicate for both extrema. Add sparse and chunked regression cases for the all-negative lower endpoint, all-positive upper endpoint, mixed signs and values outside each bound.

### 2. Moderate — an exhausted Stage 2 session accepts another round request and panics

The new [Stage 2 round guard](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/recursive/witness/handles_and_relation.rs#L552-L566) checks the next round number, pending state and claim, but never checks `round < num_rounds()`.

After the last valid [challenge transition](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/recursive/witness/handles_and_relation.rs#L578-L594), `next_round == num_rounds`, `pending == None`, and the current claim is valid. A request at that round therefore passes. The public backend [dispatch](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/mod.rs#L618-L629) validates ownership but does not reject the exhausted round.

**Static failure path:** the final challenge [clears the cached polynomial](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/sumcheck/relation_range_image/round_flow.rs#L4-L12); the extra request recomputes it, reaches [the dense-round linear-coefficient check](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/sumcheck/relation_range_image/dense_terms.rs#L198-L209), and then indexes `tau[num_rounds]` through [GruenSplitEq::current_tau](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-algebra/src/split_eq.rs#L112-L115). That index is out of bounds.

**Impact:** misuse of the new fallible public prover-session API can panic instead of returning an error. The ordinary protocol driver uses the advertised round count, so this is not evidence that ordinary proofs panic, nor a violation in verifier-reachable code.

**Repair:** reject `round >= self.num_rounds()` before invoking arithmetic, as [Stage 3 already does](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/sumcheck/stage3/mod.rs#L123-L131). Keep the completed session usable for `finish` after rejection. Extend the existing `two_round_relation_session()` fixture: run its two legal polynomial/challenge pairs, request polynomial 2 using the current claim, and assert `Err`; also exercise the public Stage 2 dispatch. This finding is supported by static call-path analysis; I did not execute that regression test.

### 3. Moderate — response-model calibration diagnostics were removed during the cutover

The feature previously emitted an accepted-response sample containing exact source energy, observed response energy, challenge-scaled conditional mean, geometry, nonce and attempt count. See the [base logging path](https://github.com/LayerZero-Labs/akita/blob/4215b5c2bec75fb2810905a6e6c57e4e79d530e6/crates/akita-prover/src/protocol/fold_grind.rs#L432-L498); the terminal path had the corresponding sample too.

At this head:

- [Admission exits before measuring response energy](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/fold.rs#L398-L410) when no security L2 cap is active, even if diagnostic collection is enabled.
- Where energy is computed, [the accepted result drops the aggregate](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/fold.rs#L734-L761).
- The protocol now [logs nonce, attempts and cap only](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-prover/src/protocol/fold_grind.rs#L295-L301). The former `conditional_mean_l2_sq`/`fold response model sample` emission is absent from Rust sources. Remaining source-moment diagnostics do not replace that sample.

**Impact:** enabling the documented calibration feature no longer supplies the observed-versus-predicted response measurements. This is an existing capability regression, and the [new specification explicitly promises its preservation](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/specs/opaque-prover-consumer.md#L433-L462), including [all five diagnostic values](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/specs/opaque-prover-consumer.md#L1350-L1355).

**Repair:** expose feature-gated aggregate diagnostics through the opaque result/handle, including accepted-response energy and the necessary source energy; derive and emit the conditional mean in protocol code. Collect diagnostics independently of whether the security policy has an L2 cap. Preserve sparse/chunked and terminal samples, and compare their fields against the baseline under a tracing subscriber. Disabled builds should still avoid the extra scans. No witness-buffer export is needed.

## Specification alignment

| Requirement | Assessment |
| --- | --- |
| Protocol crate independent of CPU representation | Substantially met: generic orchestration uses backend contracts; CPU sources, commitment material and arithmetic moved into the CPU crate. |
| Owner/setup/scope/commitment/operation binding | Substantially met in inspected paths: exact commitment identity and opening lineage are checked; proof scopes isolate mutable state. Exhausted Stage 2 progression still needs finding 2. |
| Unchanged acceptance, transcript and deterministic proof behavior | Finding 1 violates acceptance parity. Static transcript-order tracing found no other concrete ordering defect. Full independent base/head byte equality was not established by this review. Equal serialized sizes and successful round trips do not establish byte equality. |
| Diagnostic event parity | Not met; finding 3. |
| No global mutex/registry lookup in CPU round paths | Not met literally: binding validation calls `validate_context`, which consults the shared active-proof registry. See [lifecycle.rs](https://github.com/LayerZero-Labs/akita/blob/ea74e9544b69424fb32d702024edf7d2c12dd589/crates/akita-cpu-backend/src/opaque/lifecycle.rs#L241-L254). No contention regression was measured here. Consider a validated per-session lease, or explicitly revise this acceptance criterion. |
| CPU performance and memory gates | Partially established; latest measurements below leave exceptions to resolve against the specification's 3% timing/no-memory-increase criteria. |

The [current CI benchmark](https://github.com/LayerZero-Labs/akita/pull/36#issuecomment-5748252683) is for **ea74e95**, with interleaved base/head runs and three measured samples. All 13 profiles passed, and the earlier large proving regressions are mostly gone. However, Fp128 one-hot nv36 direct proving is **+3.5%**; recursive multi-group setup is **+4.8%**, and W8R2 recursive setup is **+4.5%**. Peak RSS increases are **+6.9%**, **+6.4%**, and **+4.5%** for the three recursive profiles. These are CI observations, not fresh local measurements or proven attribution to one allocator/cache. Reconcile them with the acceptance criteria or investigate and rerun the affected cases; a successful benchmark job alone does not satisfy those criteria. Prepared NTT-cache sizes and proof sizes in that report are unchanged.

The principal abstractions are earned: they establish real ownership and protocol boundaries rather than simply renaming functions.

| Component | Responsibility worth preserving |
| --- | --- |
| Generic prover + fallible `SumcheckKernel` | Transcript ownership, public claims, round ordering and error propagation. |
| `CpuBackend`, prepared setup, prefix cache | Immutable reusable setup/source/commitment resources and CPU arithmetic. |
| Proof scope + private operation binding | Lifetime, owner/setup identity, admitted schedule and operation provenance. |
| Accepted fold / recursive witness handles | Retained private coefficients, chunk-local admission, opening lineage and controlled witness transitions. |
| Stage 1/2/3 and EOR sessions | Mutable round state; only scheduled public messages cross the boundary. |

The newest setup-prefix optimization has meaningful safeguards: its in-process validation bit is private, insertion invalidates it, serialization omits it, deserialization restores the untrusted path, and imports check the setup seed. The cache issues fresh handles rather than sharing mutable proof sessions. I found no actionable untrusted-cache bypass in these inspected paths; this still relies on the documented trusted setup/backend construction boundary.

Maintenance concerns: session progression checks are duplicated and already differ at the exhausted-round boundary; keep those invariants explicit and cover them consistently. Stage 1's manual protocol loop is another transcript-sensitive maintenance surface alongside the generic sumcheck driver, though its public transition/L2 merge responsibilities justify more than a trivial adapter. The opaque layer should carry the missing diagnostics rather than force protocol code to recover raw representations.

## Components intentionally deferred or still missing

- This is a CPU ownership cutover, not an implemented GPU backend or device-residency guarantee. The external backend fixture establishes API implementability, not a second working prover or accelerator-performance result.
- The CPU backend remains trusted prover implementation code; opaque Rust handles do not make it an adversarial memory/security sandbox.
- Source-level API changes are intentional and allowed by the repository compatibility policy; moved public paths alone are not review defects.
- Independent base/head transcript and proof-byte parity, and concurrent-proof contention/memory characterization, remain unverified in this review.

## Review coverage and validation

Reviewed the full base/head inventory and architecture changes, with separate passes over ownership/admission/lifecycle, CPU folding/arithmetic/NTT resources, and protocol/sumcheck/EOR/verifier boundaries. Traced setup-prefix persistence and the newest cache changes separately. Moved arithmetic was compared selectively at its changed boundaries; this is not a fresh formal audit of every relocated arithmetic line.

No additional actionable soundness defect was established in the inspected verifier shape checks, terminal canonical decoder boundary, EOR aggregate checks, transcript sequencing, per-chunk admission/overflow checks or opening/commitment lineage. In particular, the suspected large one-hot EOR densification regression does not follow the production root path: root proving sets `run_eor=false`.

Local validation was static base/head and call-path inspection plus the executed admission-predicate counterexample. I did **not** rerun the full Cargo suite, portability suite or benchmarks, and did not edit production code. At the final head refresh, GitHub reports **41 successful checks, 2 skipped, 1 neutral, none failing or pending**. Skipped: Deploy Akita Book and one large-PR advisory instance. Neutral: Socket Security PR Alerts. Successful checks include both test shards, Clippy, transcript checks for both backends, fuzzing, Wasm, x86/AArch64 checks and all benchmark shards. Those checks are useful evidence but do not cover the counterexamples above.

**Suggested repair order:** correct fold admission and add endpoint coverage; reject exhausted Stage 2 rounds; restore diagnostic aggregates and event parity; then resolve the explicit performance/specification exceptions. This comment is a review with proposed repairs, not a code update or merge approval.

crates/akita-cpu-backend/src/opaque/mod.rs
```diff
@@ -0,0 +1,885 @@
...
+                        .into(),
+                ));
+            }
+            self.validate_binding(&opening_binding)?;
+            self.validate_binding(&material_binding)?;
+            binding.validate_lineage(&opening_binding)?;
+            binding.validate_lineage(&material_binding)?;
+            opening_binding.validate_group(group_index, opening_batch.num_groups())?;
+            material_binding.validate_group(group_index, opening_batch.num_groups())?;
+        }
```
mtbadakhshan, line 237: **Delete the second backend abstraction layer.**

Witness construction goes through `OpaqueRecursiveWitnessBuildKernel → CpuWitnessBuildKernel → RecursiveWitnessAssemblyKernel`, with the private backend contracts implemented only for `CpuBackend`. The inner build reconstructs the consumer ID and setup digest, creates a level-zero binding and empty opening bindings, and this outer adapter immediately replaces them. Required setup also travels as `Option` and is checked again at each layer.

Keep the external backend contract and implement it with concrete CPU functions and private state. Pass the validated binding into construction once and make required setup a reference. The duplicate private build/handle-family contracts and single-use implementation macros should disappear. The same simplification applies to the private Stage 1/2 and prepared-opening adapters; useful arithmetic primitives and source-polymorphism traits can stay.

crates/akita-cpu-backend/src/opaque/backend.rs
```diff
@@ -0,0 +1,396 @@
...
+use std::collections::BTreeMap;
+use std::sync::{Arc, Mutex};
+
+type SetupPrefixCacheValue = Arc<dyn Any + Send + Sync>;
+type SetupPrefixCacheCell = Arc<Mutex<Option<SetupPrefixCacheValue>>>;
+
+/// Reusable owner of CPU setup resources and independent proof scopes.
+pub struct CpuBackend {
+    identity: Arc<BackendIdentity>,
+    prepared: Box<dyn CpuSetupResources>,
```
mtbadakhshan, line 21: **Preserve the configuration type instead of erasing and recovering it.**

This backend fixes one `Cfg` at construction, then stores `TypeId`, erased setup/catalog objects, and downcast cache values. Methods reintroduce `Cfg`/`F`/`E` and dynamically reject combinations the type system could rule out. Since the prover already uses statically generic backend contracts, `CpuBackend<Cfg>` with typed setup, catalog, and prefix cache values would delete these mismatch states and checks. Keep instance-ownership checks: two backends can still share a configuration.

Prepared openings have the same problem in `opaque/recursive/witness/handles_and_relation.rs`: `retained_source` is `Option<Arc<dyn Any + Send + Sync>>`, but its only production writer and both readers use `RetainedOpeningSource<F, E>`. Store that enum directly and construct the handle completely. If terminal-native openings need different fields, represent them as an explicit variant.

crates/akita-prover/src/protocol/ring_relation.rs
```diff
@@ -778,136 +341,135 @@ impl RingRelationProver {
...
+        let witness_plan = crate::backend::ValidatedRecursiveWitnessPlan::new(
+            expected_witness_len,
+            commitment_ring_dimension,
+        );
+        let witness_handle =
+            crate::backend::OpaqueRecursiveWitnessBuildKernel::finish_recursive_witness(
+                consumer,
+                build_handle,
+                folds,
+                crate::backend::RecursiveWitnessPublicInputs {
```
mtbadakhshan, line 451: **Give the public relation instance one canonical owner.**

The protocol has just constructed and validated `RingRelationInstance`, but passes only extension degree, gamma, and row coefficients here. `opaque/witness_build/assembly.rs:838` independently constructs another instance from these values plus retained group/layout/RHS state, validates the groups again, and uses that instance to build the witness and relation plan. Group ordering, opening-family handling, and RHS assembly now have two orchestration paths to keep synchronized.

Pass a validated reference to the instance already constructed here into `finish_recursive_witness`. The CPU should check it against retained source/challenge provenance and use the same instance for witness construction. Remove the second public-instance assembly path and its intermediate output wrapper. Preserve coverage for compressed/uncompressed payloads, preceding setup-prefix groups, both opening families, and mixed-lineage rejection.

crates/akita-cpu-backend/src/opaque/lifecycle.rs
```diff
@@ -0,0 +1,404 @@
...
+}
+
+pub(crate) struct BackendIdentity {
+    backend_id: u64,
+    setup_digest: [u8; 32],
+    next_scope: AtomicU64,
+    next_operation: AtomicU64,
+    active: Mutex<HashMap<ProofScopeId, ActiveProof>>,
+}
+
```
mtbadakhshan, line 105: **Let each proof session own its state directly.**

`CpuProofSessionHandle` retains an owner pointer and numeric scope ID, while the actual plan, admitted commitments, group counts, and liveness live in this backend-wide mutex registry. Each Stage 2 round/challenge binding reaches `validate_context`, which locks the registry through `validate_scope` and then locks it again. Private witness and sumcheck state already live in directly owned handles, so the registry mainly recovers proof metadata and liveness.

Have the opaque session own an `Arc<ProofState>` with immutable admitted geometry and per-proof admission/liveness state, and let derived handles retain it. Root operations can receive the opaque session, as `Stage3Request` already does, plus level/group metadata. Keep owner and operation-lineage checks while removing the global lookup and repeated context reconstruction. This is a structural coupling concern, not a benchmarked contention claim; preserve concurrent-proof, foreign-owner, abort, double-finish, and post-finish rejection tests.

crates/akita-cpu-backend/src/opaque/witness_build.rs
```diff
@@ -0,0 +1,2 @@
+include!("witness_build/assembly.rs");
```
mtbadakhshan, line 1: **Decompose this into real modules before merging.**

These two `include!` fragments form one 1,939-line module: 1,030 lines of assembly plus 909 of finalization. `opaque/recursive/witness.rs` similarly combines 855 and 784 lines into a 1,639-line module. The lower PR already grows `backend/recursive/witness.rs` from 944 to 1,018 lines. Textual splitting preserves the shared namespace, imports, private fields, and coupling.

`assembly.rs` currently combines retained commitment material, prepared opening representation, decomposition, two backend-adapter layers, and relation payload construction. Delete the redundant layers first, then make commitment material, openings, witness assembly/finalization, and stage sessions ordinary modules with explicit inputs and outputs. Splitting files should reduce the concepts and invariants held together, not only satisfy a physical line-count gate.

mtbadakhshan (commented):
Architecture review of #36 stacked on #31. The dependency direction and source ownership are sound improvements. I would address the five structural findings below before merging: remove duplicate internal backend layers, retain known types, construct the public relation once, let sessions own proof state, and introduce real module boundaries.

Reviewed head: `2985d9ab`. Local validation: formatting and 84 Python repository tests passed. Rust compilation, Clippy, full protocol tests, and benchmarks were not run for this architecture review.

quangvdao:
## PR Review findings

Thank you, @RadNi, for the substantial work on this refactor and for carrying the ownership change through the application APIs, setup-prefix persistence, diagnostics, tests, and CI. The reusable owning-backend direction is valuable, and the provenance and lifecycle checks deserve to be preserved during the remaining cleanup.

**Recommendation: address the findings below before merging this reviewed head.** There is a confirmed documentation-build failure, an error-path cache-retention defect, an opening-handle contract mismatch, and substantial architecture/validation issues. Four architectural findings from the earlier review remain present. The earlier per-round registry-locking criticism is **partially addressed** and should not be repeated unchanged.

**Reviewed surface:**

```text
Base: 4215b5c2bec75fb2810905a6e6c57e4e79d530e6
Head: cfd3f4a00390f141fb129d27f088264ae733c302
```

The PR still pointed to this head when this comment was prepared. The review concerns the 409-file, 13-commit base-to-head surface, not merely the latest commit.

**Validation limits:** This review used static source/diff inspection, specification and prior-review comparison, and inspection of existing CI. No Rust compilation, new Rust regression tests, Miri/sanitizer runs, or benchmarks were executed locally for this review. It is not an independently executed, exhaustive audit of every arithmetic kernel. No proof-forgery vulnerability, transcript-binding bypass, or memory-safety defect was established; that is not proof that those classes of bugs are absent.

### 1. Moderate — Public documentation links to a private representation and fails CI

**Evidence:** [`sources/onehot/ops.rs:43`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/sources/onehot/ops.rs#L43), in the public documentation for `OneHotBatchView`:

```rust
/// `D` is the kernel dispatch dimension, as in [`OneHotView`].
```

`OneHotView` is private in the public documentation surface. The [Documentation job](https://github.com/LayerZero-Labs/akita/actions/runs/35689697996/job/106623855897) fails under warnings-as-errors with:

```text
error: public documentation for `OneHotBatchView` links to private item `OneHotView`
--> crates/akita-cpu-backend/src/sources/onehot/ops.rs:43:51
error: could not document `akita-cpu-backend`
Process completed with exit code 101.
```

**Required correction:** Explain the dimension directly, for example:

```rust
/// `D` is the ring dimension selected for this kernel invocation.
```

Alternatively, link an appropriate public abstraction. Do not re-export private view machinery just to repair documentation, and do not suppress the lint.

**Validation:** Run the exact failing documentation configuration:

```bash
RUSTFLAGS="-D warnings" RUSTDOCFLAGS="-D warnings" \
cargo doc -q --release --no-deps --no-default-features \
  --features parallel,disk-persistence,transcript-blake2b
```

This is a confirmed CI/release issue, not a cryptographic vulnerability.

### 2. Moderate — Failed setup-prefix derivations leave permanent backend-cache entries

**Evidence:** [`opaque/backend.rs::memoized_setup_prefix`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/backend.rs) inserts a per-ID cell before executing the fallible derivation:

```rust
cache
    .entry(id.clone())
    .or_insert_with(|| Arc::new(Mutex::new(None)))
```

Later:

```rust
let value = Arc::new(derive()?);
```

If `derive()` fails, the newly registered empty cell is not removed. Its key, map entry, `Arc`, and mutex remain attached to the backend.

This is reachable through the public prefix preparation/export paths: [`setup_prefix_material`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/owned_prefix.rs) calls the memoizer, while prefix profile, geometry, and capacity rejection can occur inside [`commit_setup_prefix`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/setup/setup_prefix.rs).

**Concrete failure mode:** A long-lived backend retains bookkeeping for distinct failed prefix IDs despite producing no cached material. Repeating the same ID retains only one entry. `trim_caches()` delegates to setup-transform trimming and does not remove prefix-map entries. Its documentation specifically concerns transforms, so the defect is failed-request retention, not a claim that `trim_caches()` violates its documented contract.

This is a resource-management/availability finding established by control-flow inspection, not a locally executed reproduction or a demonstrated remotely exploitable denial of service.

**Required correction:** Validate inexpensive ID/profile/capacity conditions before registering a cache cell. For remaining fallible work, publish only successful values or clean up failed in-flight cells with correct concurrency semantics. Cleanup must not remove a successful replacement or another generation's cell; preserve same-ID coordination and check cell identity where needed. Caching `Result<T, Error>` indefinitely would still retain every failed key.

**Regression coverage:** Many distinct invalid IDs must not grow retained entries; failure followed by successful retry must work; concurrent successful requests for one ID must preserve intended single-flight behavior; failure cleanup must not discard concurrent successful publication.

### 3. Moderate — The prepared-opening handle is cloneable despite the specified linear-handle contract

**Evidence:** [`CpuPreparedOpeningHandle`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/recursive/witness/handles_and_relation.rs) has a public `Clone` implementation:

```rust
#[doc(hidden)]
#[derive(Clone)]
pub struct CpuPreparedOpeningHandle<F: Field, E: Field> {
    // ...
}
```

Cloning preserves its operation binding; `#[doc(hidden)]` does not hide the trait implementation. The [proposed opaque-consumer specification](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/specs/opaque-prover-consumer.md) says linear handles, including openings, do not implement `Clone` or `Copy`.

The [`terminal_native_witness_opening`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/mod.rs) operation takes its handle by value, but validates binding/terminal state rather than globally unique consumption. A caller can duplicate the handle before this consuming operation.

**Qualification:** The specification is proposed and has other stale details. The established finding is implementation/contract disagreement and loss of the apparent consume-once property—not a demonstrated soundness or secrecy break. Private fields are not exposed by cloning.

**Required correction:** Remove public `Clone` from the prepared-opening handle. Keep cheap internal sharing of immutable storage, and use an explicit private snapshot operation only where necessary. Fold retries should borrow the prepared opening. Intentionally reusable immutable source/commitment handles should remain cloneable. If reusable prepared openings are the intended design, explicitly revise the contract and consuming API semantics instead of leaving them contradictory.

**Regression coverage:** A compile-fail test should prevent the public prepared-opening associated type from satisfying `Clone`; repeated fold probes through a borrowed opening should continue to work.

Findings 4–7 also appear in the [earlier architecture review](https://github.com/LayerZero-Labs/akita/pull/36#pullrequestreview-5271707793). They were rechecked against the head above rather than treated as new vulnerabilities. Finding 8 records the partial improvement since that review.

### 4. Moderate — The public relation instance has two independent construction paths

**Evidence:** [`akita-prover/src/protocol/ring_relation.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-prover/src/protocol/ring_relation.rs) constructs and validates `RingRelationInstance` from public group openings, extension degree, layout, batching coefficients, RHS, and dimensions. It then sends only extension degree, gamma, and row-coefficient rings into `finish_recursive_witness`.

[`opaque/witness_build/assembly.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/witness_build/assembly.rs) independently constructs another instance from those inputs plus retained openings, layout, and RHS, validates public groups again, and uses this second instance for witness construction.

**Why it matters:** Group order, setup-prefix groups, opening-family handling, and compressed/uncompressed RHS assembly now have two orchestration paths to keep synchronized. This duplicates public-statement authority across the intended protocol/private-backend boundary. No current supported-input mismatch between the two instances was established; this is a security-sensitive maintenance defect, not a claimed present proof forgery.

**Required correction:** Construct the public instance once in protocol code and pass a validated reference into witness finalization. The CPU should check agreement with retained source, commitment, opening, and challenge provenance and build the witness against that same instance. Remove the duplicate assembly path and the intermediate output carrier returning the duplicate. Adding another equality check while permanently preserving both constructors is not the canonical fix.

**Regression coverage:** Both opening families, compressed/uncompressed payloads, preceding setup-prefix groups, heterogeneous groups, and deliberately mixed opening/fold/commitment lineage.

### 5. Moderate — A redundant private CPU backend abstraction stack creates partially initialized state

Witness construction still follows:

```text
OpaqueRecursiveWitnessBuildKernel
    → CpuWitnessBuildKernel
        → RecursiveWitnessAssemblyKernel
```

**Evidence:** The inner adapter in [`assembly.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/witness_build/assembly.rs) reconstructs the consumer ID from the scope, recomputes the setup digest, creates a level-zero binding, and initializes an empty opening-binding collection. The outer [`opaque/mod.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/mod.rs) adapter already has validated context and replaces that bookkeeping. Required setup travels as `Option<&PreparedSetup>` through layers that immediately reject `None`, with single-backend implementation macros further obscuring the call graph.

**Why it matters:** This adds initialization states and identity-derivation responsibilities without a genuine additional implementation choice. A reviewer must prove that temporary or unscoped internal state is always rebound before use.

**Required correction:** Keep the public opaque contract, but implement it with concrete private CPU functions. Pass the validated binding and required setup reference directly into construction and produce a fully initialized handle once. Delete private build/handle-family interfaces that only re-describe this CPU implementation. Preserve substantive arithmetic and source-polymorphism traits.

**Validation:** Preserve foreign-owner, wrong-scope/level/group, and mixed-operation rejection tests, and eliminate ordinary production construction through temporary level-zero or unscoped bindings.

### 6. Moderate — Known configuration and opening-source types are erased and then recovered dynamically

**Evidence:** [`CpuBackend`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/backend.rs) fixes one `Cfg` at construction but retains a configuration `TypeId`, erased setup/catalog objects, and erased prefix-cache values. Methods reintroduce type parameters and downcast or dynamically reject combinations.

[`CpuPreparedOpeningHandle`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/recursive/witness/handles_and_relation.rs) similarly stores:

```rust
retained_source: Option<Arc<dyn Any + Send + Sync>>
```

The production writer/readers use the concrete `RetainedOpeningSource<F, E>` family, yet construction permits a missing-source intermediate state and recovery requires a downcast.

**Why it matters:** Information already known is discarded, then restored through additional runtime states and checks in a field/ownership-sensitive implementation.

**Required correction:** Prefer `CpuBackend<Cfg>` with typed setup, catalog, and cache values. Store the concrete retained-source enum in openings and represent terminal-native openings as an explicit variant if their state differs. Localize any genuinely required application-level erasure outside this implementation.

Keep runtime **instance-ownership** checks: two backends with the same configuration remain distinct owners. Static configuration typing does not replace source provenance, proof-scope identity, or operation-lineage checks.

### 7. Moderate — Textual file splitting preserves the mixed-responsibility modules

**Evidence:** [`opaque/witness_build.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/witness_build.rs) is:

```rust
include!("witness_build/assembly.rs");
include!("witness_build/finalize.rs");
```

[`opaque/recursive/witness.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/recursive/witness.rs) likewise textually includes `handles_and_relation.rs` and `opening_and_flat.rs`.

The issue is not a line-count threshold. Assembly still combines retained commitment material, terminal publication, prepared openings, decomposition, backend adapters, and public relation construction in one shared namespace. Physical splitting has not assigned clear owners to those private invariants.

**Required correction:** Remove the duplicate layers first, then use ordinary modules for commitment material, prepared openings, witness assembly/finalization, and stage sessions, with explicit imports and narrow visibility. Do not restore the old coupling by making every item `pub(crate)`. Do not reuse the earlier review's exact combined line counts as current measurements without recounting them.

### 8. Residual architecture concern — Proof-state ownership remains divided, but round-loop locking is improved

The earlier criticism must be narrowed. Current Stage 1/2, EOR, and Stage 3 sessions retain a `ScopeLease` and use leased validation during progression. The assertion that each Stage 2 round still repeatedly locks the global registry is no longer accurate for these paths: see [`opaque/mod.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/mod.rs), [`eor.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/eor.rs), and [`stage3/mod.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/sumcheck/stage3/mod.rs).

The remaining concern is that [`lifecycle.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/lifecycle.rs) keeps admitted plans and commitment bookkeeping in a backend-wide registry while computational state, bindings, and leases live in derived handles.

**Suggested direction:** Have the proof session own `Arc<ProofState>` containing immutable admitted geometry and per-proof admission/liveness state; derived handles can retain it. Preserve owner and operation identity checks while removing unnecessary lookup/context reconstruction.

This is less urgent than duplicate public-instance construction. There is no benchmark evidence here of current contention or a measured performance regression. Preserve concurrency, foreign-owner, abort, double-finish, and post-finish rejection coverage.

## Validation and specification gaps

### 9. Moderate validation gap — The external fixture does not execute successful alternate-backend orchestration

**Evidence:** [`akita-prover/tests/external_backend.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-prover/tests/external_backend.rs) type-checks the actual generic entrypoint with a non-CPU backend. This is useful. However, `begin_proof` rejects, the operation implementations are rejecting stubs, and the helper invoking `batched_prove` is not called by the fixture's tests. Those tests cover message/metadata construction and a separate lifecycle guard.

This establishes external implementability of the signatures, not successful handoff through openings, witness construction, stages, terminal publication, and cleanup for an alternate handle family. The real [`backend_ownership.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-pcs/tests/backend_ownership.rs) CPU integration test exercises successful proofs, but that is a different property.

**Suggested fix:** Keep the CPU-independent compile fixture. Add a successful integration adapter using distinct newtype handles around a CPU-backed implementation, in an integration-test crate allowed to depend on both sides. Instrument operation ordering/lifecycle, cover recursion, EOR, setup-prefix/Stage 3, and terminal publication, and inject failures after admission and during later stages to test cleanup and subsequent independent proofs. A second cryptographic implementation is not necessary.

### 10. Low-to-Moderate — The proposed specification still contains superseded ownership/API designs

**Evidence:** [`specs/opaque-prover-consumer.md`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/specs/opaque-prover-consumer.md) retains an older module tree inside the prover crate, borrowed-source ingestion, and numeric-scope lifecycle examples, while the implementation uses a separate owning CPU crate, consumed source storage, and opaque session authority. It remains marked proposed.

**Required correction:** Bring the specification into agreement with the actual crate boundary, owning import, session authority, and reuse/consumption rules. Record rejected designs separately rather than interleaving them with intended end-state contracts. In particular, resolve opening-handle linearity explicitly. This is documentation/contract repair, not a reason to retroactively declare every implementation detail intentional.

### Proof-byte preservation requires separate evidence from current determinism

The new ownership test compares concurrent and subsequent proofs from the same current implementation. It supports current determinism, not independently verified base/head byte equality.

No base-versus-head differential test was executed for this review. Thus, preservation of proof bytes and transcript order is **not fully independently verified here**, rather than a confirmed regression. Add or provide fixed-input comparisons of serialized proofs and transcript events against the pinned base across both opening families, small-field EOR, setup-prefix recursion, and multi-group cases. A second invocation of the same new implementation is not an independent parity oracle.

## Checks that did not become defects

These protections should survive the simplification:

- **Exact source and computation provenance:** Witness assembly checks retained public commitment material and exact committed-source identity; accepted-fold checks bind to the opening operation rather than shape alone. No same-shape substitution bug was established in the inspected paths. See [witness admission/assembly](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/mod.rs) and [opening/fold checks](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/openings.rs).
- **Cross-owner commitment import:** [`import_commitment`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/opaque/owned_commit.rs) recomputes from the retained source under the receiving setup and compares public commitment and retained material before issuing a fresh handle.
- **Setup-prefix provenance:** [`SetupPrefixProverRegistry`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/commitment/setup_prefix.rs) does not serialize in-process validation authority; deserialization starts unvalidated and public insertion invalidates the flag. No obvious mutation/serialization bypass was found.
- **Guard cleanup:** [`ProofScope::finish`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-prover/src/backend/scope.rs) marks completion only after backend success; failure still triggers abort on drop. The external fixture covers that guard behavior.
- **EOR public checks:** [`opening_reduction.rs`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-prover/src/protocol/prove/opening_reduction.rs) compares expected claims, reconstructs aggregate input from public partials, checks the final weighted claim, and derives transparent factors outside the private session.
- **Terminal payload validation:** [`build_terminal_response_from_payload`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-types/src/proof/tail_segments.rs) checks geometry and invokes terminal Golomb-payload validation rather than accepting arbitrary encoded bytes.
- **Snapshots are not full-witness-copy evidence:** [`PackedSignedDigits`](https://github.com/LayerZero-Labs/akita/blob/cfd3f4a00390f141fb129d27f088264ae733c302/crates/akita-cpu-backend/src/sources/packed_digits/mod.rs) is `Arc`-backed. An internal snapshot clone does not imply a full coefficient-allocation copy, so no such performance finding is asserted.

The stated boundary is API opacity, not in-process secrecy. Ordinary private CPU views or storage are not vulnerabilities merely because the backend can access them.

## Specification alignment

| Intended property | Assessment |
|---|---|
| Public orchestration owned by protocol; private computation by backend | Correct direction, but public relation construction is duplicated across the boundary (4). |
| Exact source, commitment, and operation provenance | Meaningful checks are present in inspected paths; preserve them. |
| Linear transient handles | Prepared-opening `Clone` contradicts the proposed contract (3). |
| Reusable immutable commitments and independent mutable sessions | Covered by successful concurrent CPU proof integration tests. |
| Cleanup after rejection | Guard behavior is correct in the inspected path; failed prefix-cache registration leaks retained bookkeeping (2). |
| External backend support | Signature implementability is covered; successful alternate-handle execution is not (9). |
| Determinism and byte preservation | Current determinism is tested; base/head parity was not independently established here. |
| Clear implementation ownership/module boundaries | Erasure, redundant private traits, and textual splitting remain (4–8). |
| Accurate developer contract | Proposed specification still mixes superseded and current designs (10). |

## CI status and validation limits

In inspected [CI run 35689697996](https://github.com/LayerZero-Labs/akita/actions/runs/35689697996), both test shards, all three Clippy configurations and dependency hygiene, Keccak/Blake2b transcript semantics, schedule-artifact checks, formatting, Rust file caps, TOML formatting, dependency checks, and spelling checks passed. **Documentation failed as described in finding 1.** The documentation log checks out merge commit `7f504c8bad094197b90add0662ae93368a252bb1`, merging the reviewed head into the reviewed base.

These are observed existing CI results, not locally executed review validation. The PR body reports 1,844 passing implementation tests; that count was not independently rerun or recounted for this review. No contention/performance measurement is claimed, and the cache finding is a static error-path proof rather than an executed Rust reproduction.

## Suggested repair order

1. **Repair definite failure paths:** Fix the rustdoc link, prevent failed prefix derivations from leaving persistent cache registrations, add the negative-cache tests, and resolve prepared-opening clone/consumption semantics.
2. **Simplify the statement/witness boundary:** Construct `RingRelationInstance` once; pass validated identity and required setup directly into concrete CPU construction; delete redundant build/assembly layers.
3. **Make remaining state explicit:** Preserve known types, introduce real modules, update the specification, and add successful alternate-handle integration and pinned base/head parity coverage. Consider the remaining proof-registry ownership cleanup without describing the already-improved round paths as a current contention bug.

The strongest correction is to delete duplicated states and responsibilities, not add more wrappers or blanket revalidation. Thank you again for the work on the owning backend and its integration coverage; retaining those improvements while simplifying these boundaries would make the change substantially easier to trust and maintain.

quangvdao:
Following up on the RingRelation finding (the one about a single owner for `RingRelationInstance`). @RadNi, you said on Slack that some backends may want the relation instance and some may not, so it's better not to enforce it. I don't think that objection applies to what the finding asks for, for three reasons.

**1. The protocol already builds the instance for every backend.** [`ring_relation.rs:427`](https://github.com/LayerZero-Labs/akita/blob/dd1cf825c961918fc4829839a3c58d6feaa8f60c/crates/akita-prover/src/protocol/ring_relation.rs#L427-L441) constructs `RingRelationInstance::new(...)` and runs `validate_prepared_relation_groups` whichever backend is in use. The only thing that depends on the backend is what gets passed to `finish_recursive_witness` at [`:447`](https://github.com/LayerZero-Labs/akita/blob/dd1cf825c961918fc4829839a3c58d6feaa8f60c/crates/akita-prover/src/protocol/ring_relation.rs#L447-L457). Today that is `RecursiveWitnessPublicInputs { extension_degree, gamma, row_coefficient_rings }`, which is three of the instance's seven inputs.

**2. The CPU backend then builds a second instance.** [`assembly.rs:227`](https://github.com/LayerZero-Labs/akita/blob/dd1cf825c961918fc4829839a3c58d6feaa8f60c/crates/akita-cpu-backend/src/opaque/witness_build/assembly.rs#L226-L241) calls `RingRelationInstance::new` again. It uses its own `relation_group_openings` and its own `relation_rhs`, which it assembled from private build state. It then runs `validate_prepared_relation_groups` a second time. Nothing checks that the two instances are equal. They agree only because two separate code paths compute the same group openings and right-hand side. If those paths diverge, the prover emits a witness for a relation the transcript never bound, and the verifier rejects. That makes it a completeness failure rather than a soundness failure. It is still two owners for one public object.

**3. Other backend entry points already receive `&RingRelationInstance`.** Both [`backend/plans/relation.rs:8`](https://github.com/LayerZero-Labs/akita/blob/dd1cf825c961918fc4829839a3c58d6feaa8f60c/crates/akita-prover/src/backend/plans/relation.rs#L8) (Stage 2) and [`backend/stage3.rs:17`](https://github.com/LayerZero-Labs/akita/blob/dd1cf825c961918fc4829839a3c58d6feaa8f60c/crates/akita-prover/src/backend/stage3.rs#L17) (Stage 3) take `relation: &'a RingRelationInstance<F>`. The PR already requires every backend to accept the instance twice. Passing it once more, to witness finishing, adds no new requirement.

**Proposal.** Replace `RecursiveWitnessPublicInputs` with `relation: &RingRelationInstance<F>` in `finish_recursive_witness`.
- A backend that doesn't need the instance ignores the argument, so nothing is enforced on it.
- The CPU backend drops its `RingRelationInstance::new` call and its second `validate_prepared_relation_groups` call.
- If the CPU backend wants a defensive check, it compares its private `relation_rhs` against `relation`'s, and does not build a parallel instance.
- The witness-side data (`RingRelationWitness::from_groups`) stays backend-private, as it is today.

I haven't made this change on the branch, since it's a design call on your PR. I'm happy to implement it if you agree.

---

Separately, I synced this branch with #31 (which now includes `main`) and pushed as a fast-forward from `2438282a4`:
- `8067ca877`: merge of #31 into this branch. It resolves 16 conflicts, including the `with_ring_switch_cache_limit` port from #35, #38's `MAX_I8_LOG_BASIS` gate and signed-boundary test, and #31's challenge-count checks for one-hot and dense scalar/batch decompose-fold. Those checks are ported onto `decompose_sources` in `sources/onehot/decompose_fold.rs`.
- `9e5eee501`: `profile/akita-recursion` builds against the owning `CpuBackend<C>`. The Jolt verifier profile smoke workflow compiles this code outside the workspace.
- `ad4589b27`: `custom_source::aggregate_decompose_fold_witnesses` is public again. It restores what #31's external batch-kernel test relied on, and adds an integration test in `crates/akita-cpu-backend/tests/`.
- `dd1cf825c`: the book and `docs/compute-backends.md` examples now use `CpuBackend::<Cfg>::new(...)`, `import_source(...)` and `commit(...)`, replacing the old turbofish forms.

Local validation: all AGENTS.md preflight gates (including schedule regeneration with no artifact diff and doc guardrails), the three CI clippy configurations, the CI nextest command (1871 passed), and the Jolt verifier profile smoke commands (`akita-recursion-glue` tests with and without `trusted-benchmark-artifact`, plus `cargo check --release`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/akita-cpu-backend/src/opaque/mod.rs
mtbadakhshan: Re-reviewed at `f875eb55`: **partially addressed**. The private witness-build layers are gone, and construction now receives the validated bindings directly. That part is fixed.

The remaining Stage 1/2 adapters still preserve the duplicated backend interface. For example, [`RecursiveWitnessStage1Kernel`](https://github.com/LayerZero-Labs/akita/blob/f875eb55e4914eaddb0273d450c1ba4f53a24f98/crates/akita-cpu-backend/src/opaque/consumer_kernels.rs#L42) repeats the public Stage 1 operations, and [its implementation](https://github.com/LayerZero-Labs/akita/blob/f875eb55e4914eaddb0273d450c1ba4f53a24f98/crates/akita-cpu-backend/src/opaque/recursive/witness/stage_sessions.rs#L457) forwards round/challenge/transition/finish calls directly to `DigitRangeSession`. Its constructor even requires `Some(prepared)` without using the setup.

Please finish this simplification: keep the public backend contracts, perform binding and policy validation there, and call concrete CPU session methods or constructors directly. Remove the redundant private Stage 1/2 contracts and setup parameters that carry no computation. The witness-build fix demonstrates the intended structure; this thread remains open for the remaining adapters.

crates/akita-cpu-backend/src/opaque/backend.rs
mtbadakhshan: Re-reviewed at `f875eb55`: **mostly addressed**. `CpuBackend<Cfg>`, typed setup/catalog/cache values, and the explicit `PreparedOpeningSource` variants resolve the type-erasure and incomplete-opening-construction concerns.

The remaining gap is the extension-field relationship. [`validate_extension<E>`](https://github.com/LayerZero-Labs/akita/blob/f875eb55e4914eaddb0273d450c1ba4f53a24f98/crates/akita-cpu-backend/src/opaque/backend.rs#L177) still compares `TypeId<E>` with `TypeId<Cfg::ExtField>`, while [the Stage 1 implementation](https://github.com/LayerZero-Labs/akita/blob/f875eb55e4914eaddb0273d450c1ba4f53a24f98/crates/akita-cpu-backend/src/opaque/mod.rs#L350) constrains only `Cfg::Field = F` and independently introduces `E`. Invalid backend/extension combinations therefore still compile and are rejected repeatedly at runtime.

Please constrain the relevant implementations to `Cfg: CommitmentConfig<Field = F, ExtField = E>` (or use the associated extension type directly), propagate that relationship through the handles, and remove the runtime extension-type checks. Keep backend-instance and lineage checks; those validate different invariants.

crates/akita-cpu-backend/src/opaque/lifecycle.rs
mtbadakhshan: Re-reviewed at `f875eb55`: **partially addressed**. The new `ScopeLease` fixes the repeated registry lookup in Stage 1/2 round and challenge operations. The original hot-path observation no longer applies to those operations.

The ownership concern remains: [`CpuProofSessionHandle`](https://github.com/LayerZero-Labs/akita/blob/f875eb55e4914eaddb0273d450c1ba4f53a24f98/crates/akita-cpu-backend/src/opaque/lifecycle.rs#L11) still contains only an owner pointer and numeric scope ID, while [`ActiveProof`](https://github.com/LayerZero-Labs/akita/blob/f875eb55e4914eaddb0273d450c1ba4f53a24f98/crates/akita-cpu-backend/src/opaque/lifecycle.rs#L108) keeps the plan and admitted commitments behind the backend-wide `Mutex<HashMap<ProofScopeId, ActiveProof>>`. `proof_plan`, commitment admission, and commitment validation still recover per-proof state through that registry.

Please move the immutable plan/geometry and per-proof admission/liveness state into an `Arc<ProofState>` owned by the proof session and retained by derived handles. Pass the opaque session to root operations that need that state, retaining owner/lineage checks and explicit invalidation when the owning session ends. This addresses the remaining ownership coupling; it is not a claim of measured contention. Preserve concurrent-proof, foreign-owner, abort, and post-finish rejection coverage.

mtbadakhshan (approved):

PR #40, opened 2026-09-21
fix(planner)!: derive all root sizing from producer contracts

quangvdao:
## Summary

This PR makes `CommittedSourceContract`—source class plus declared coefficient bound—the canonical input to every root-planning calculation.

Before this change, planner entry points carried only `HonestFoldPolicySpec`. That preserved whether a producer was balanced or one-hot, but lost the producer's bound for precommitted groups. Consequently, honest-fold sizing could overprice balanced digit planes, selective-L2 response sizing treated bounded precommitted producers as full-field sources, and the two paths maintained separate interpretations of the same source.

This PR derives both models from `CommittedSourceContract::source_norms`, propagates the complete producer declaration beside every frozen precommit descriptor, and rejects generation requests that assign conflicting contracts to the same catalog lookup geometry.

The verifier and proof format are unchanged. Planning freezes the resulting response caps and SIS ranks into the schedule; verification continues to enforce those frozen values.

### Diff metadata

- Base: `63c81e0d9fa9c29d5d7e25f636413196487e3427` (`codex/planner-candidate-reuse`, PR #39)
- Head: `cc487a20825cf2b8f9bc879a1d845ef56fb4a28b`
- Commits: 7
- Files changed: 27
- Diff: 552 insertions, 484 deletions
- Specification: no new design spec; the validated `CommittedSourceContract` API is the source of truth, and the planner README documents the public call shape

## Motivation

A frozen `GroupCommitPhaseParams` records commitment geometry, but it does not authenticate the source class or the producer's declared coefficient bound. Those facts belong to the producer configuration.

Previously, root planning accepted a final and per-precommit `HonestFoldPolicySpec`. That was enough to distinguish balanced and unit-one-hot sources, but not to recover a bounded balanced producer:

- final-group response energy could use the planner decomposition's bound;
- precommitted balanced groups fell back to the full field width because their producer bound was unavailable;
- honest-fold witness norms were derived separately from the fold policy and selected basis;
- equivalence classes could treat producers with the same frozen geometry and fold policy—but different declared bounds—as interchangeable.

The result was conservative but potentially material overpricing of bounded precommits, plus duplicate security-sensitive source models.

## Data flow

```text
producer configuration
    -> CommittedSourceContract { class, log_commit_bound }
    -> planner source_norms(exact selected geometry)
       -> honest-fold digit and A-rank sizing
       -> selective-L2 response sizing
    -> frozen schedule caps and SIS ranks
    -> verifier enforcement
```

[...]

PR #41, opened 2026-09-21
perf(planner): reuse prepared root producers across inner bases

quangvdao:
## Summary

This PR removes repeated precommitted-producer preparation from grouped root schedule search.

For each root opening basis and opening-work item, the planner now prepares the precommitted groups and their aggregate D width once, then reuses that immutable result across every candidate inner basis. Inner-basis-dependent final-group construction, pruning, witness sizing, and candidate ordering remain unchanged.

This is stacked on #40 (`codex/preserve-precommitted-bounds`), which is stacked on #39.

### Diff metadata

- Base: `cc487a20825cf2b8f9bc879a1d845ef56fb4a28b` (`codex/preserve-precommitted-bounds`, PR #40)
- Head: `c0766b4fe7fc181d523d4db49cd89ecd67ecbf37`
- Commits: 12
- Files changed: 3
- Diff: 203 insertions, 68 deletions

The commit history includes superseded experiments and their reverts. The current base-to-head diff above is the review surface.

## Motivation

Grouped root search visits the Cartesian product of candidate inner bases, opening-work items, and candidate root layouts.

Previously, every inner-basis iteration repeated producer work that does not depend on the inner basis:

- validating root dimensions and the opening;
- deriving equivalence classes for interchangeable precommitted groups;
- validating the precommitted opening assignment;
- materializing and admitting each precommitted group for the selected opening basis;
- canonicalizing interchangeable groups; and
- summing their checked D-segment widths.

Only final-group candidate construction depends on the candidate inner basis.

## Data flow

```text
root key + source contracts + opening work + opening basis
    -> PreparedRootProducers { groups, d_width }
    -> each candidate inner basis
    -> final-group layouts, pruning, and witness sizing
```

`PreparedRootProducers` is local to one candidate-generation call. It is prepared once per opening-work item for the current opening basis and reused only across that call's inner-basis range. There is no global cache, cross-schedule cache, persistence change, or new cache identity.

## Change-surface overview

| Area | Before | After |
|---|---|---|
| Producer preparation | Recomputed for every `(inner basis, opening work)` pair | Computed once per opening-work item and opening basis |
| Inner-basis search | Builds final-group layouts for each inner basis | Unchanged |
| Candidate visitation | Inner-basis-first, then opening-work order | Unchanged |
| Unsupported producer work | Rejected independently during each inner-basis iteration | Prepared as `None` once and skipped for every inner basis |
| API surface | One crate-private helper performed preparation and candidate construction | Crate-private preparation and consumption phases; combined helper retained for tests |
| Protocol surface | Existing schedules, descriptors, and source contracts | No wire, transcript, descriptor, source-bound, or objective change |

### Prepared producer state

`PreparedRootProducers::prepare` owns the inner-basis-independent work. It retains the existing checks for:

- role-dimension projection;
- opening compatibility;
- a nonempty reduced-variable domain;
- source-contract and precommitted-group correspondence;
- precommitted opening count;
- common opening-method family;
- shared opening-ring-dimension support;
- per-group source-bound admission;
- canonical group ordering; and
- checked aggregate D-width arithmetic.

Its fields remain private and contain only the materialized precommitted groups and their aggregate D width.

### Candidate generation

`root_level_candidates_for_prepared_producers` consumes the prepared groups while recomputing every inner-basis-dependent value:

- source decomposition width;
- recursive split domain;
- commitment-slice candidates;
- final-group source norms and commitment parameters;
- local slice pruning; and
- next-witness sizing.

Both root generation paths in `CandidateDomain` build a preparation vector before entering the inner-basis loop and pair each result with the same `opening_work` entry.

### Test coverage

The existing grouped coefficient-packing candidate test now prepares one grouped producer assignment, reuses it at the next candidate inner basis, and compares the complete ordered candidate vector against fresh preparation at that same basis.

The rest of the planner suite continues to cover unsupported opening combinations, exhaustive grouped-opening domains, adaptation, totality, and complete schedule selection. No standalone benchmark harness or broad test matrix was added to this PR.

## Correctness and compatibility

- Preparation excludes `candidate_log_basis_inner`; all inner-basis-dependent calculations remain in candidate generation.
- Successful candidate visitation remains inner-basis-first and preserves opening-work and per-layout ordering.
- The focused regression compares complete candidate values and order, not only the selected candidate.
- The preparation path preserves its prior internal validation sequence.
- Preparation for all opening-work items now occurs before candidate generation begins. Malformed inputs can therefore surface a later opening-work preparation error before an earlier work item generates candidates. Valid planner inputs and emitted candidate order are unaffected; this PR does not claim identical global error interleaving.
- The change is planner-only. It does not alter verifier behavior, serialization, transcript binding, persisted data, public Rust APIs, source contracts, or the planner objective.

The mechanical reduction is from `inner_basis_count × opening_work_count` producer preparations to `opening_work_count` preparations per opening basis. This PR has no committed end-to-end benchmark and makes no numeric speedup claim.

## Deliberately excluded work

Earlier revisions also contained two independent optimizations:

- allocation-free relation-row sizing in `akita-types`; and
- canonical scalar relation-geometry construction across `akita-types` and `akita-schedules`.

Neither is required for producer reuse. They cross different ownership and API boundaries and need their own measurements and differential coverage, so they were removed to keep this PR to one optimization and one review boundary. They were not removed because they were known to be incorrect and can return as separate PRs.

The temporary Aerie timing probe was removed with them. The earlier combined 13.3–14.3% timing result therefore does not describe this final diff and is intentionally not claimed here.

## Validation at `c0766b4fe7fc181d523d4db49cd89ecd67ecbf37`

Local validation at this head completed successfully:

- all 130 planner library tests with `catalog-gen`;
- the focused prepared-versus-fresh ordered candidate comparison;
- grouped-root and adapted-schedule regression tests;
- formatting and planner all-target checks; and
- all three repository-prescribed Clippy feature configurations.

An independent review found no correctness issue in the minimized diff.

At publication time, GitHub reports 43 successful checks, three in progress, one skipped deployment check, and no failures. The running checks are the fp128 profile benchmark, schedule artifact drift, and Keccak transcript semantics.

## Reviewer map

Suggested order:

1. `crates/akita-planner/src/planner.rs`
   - `PreparedRootProducers::prepare`
   - `root_level_candidates_for_prepared_producers`
2. `crates/akita-planner/src/schedule_params/suffix_dp/candidates.rs`
   - `prepare_root_work`
   - the batched and streamed root candidate-generation paths
3. `crates/akita-planner/src/test/schedule_params_candidate.rs`
   - exact reused-versus-fresh candidate comparison


omibo:
## Inline comment 1

**File:** `crates/akita-planner/src/schedule_params/suffix_dp/candidates.rs`  
**Line:** 710  
**Title:** `[P2] Delete the unreachable root-generation branch`

`derive_selected_suffix_schedule` calls `visit_root_batches` whenever `root_level_key.is_some()` and calls `generate_for_opening_basis` only in the `else` branch. Therefore `root_preparations` is always `None` here, and the root-specific block through line 777 is unreachable in production. Please remove this branch and make `generate_for_opening_basis` recursive-only. Keeping this dead duplicate path makes the control flow misleading and contributes to growing an already oversized file.

## Inline comment 2

**File:** `crates/akita-planner/src/schedule_params/suffix_dp/candidates.rs`  
**Line:** 924  
**Title:** `[P2] Keep root preparation bounded by work item`

`visit_root_batches` promises bounded temporary ownership, but `prepare_root_work` eagerly materializes a `PreparedRootProducers`—including a `Vec<GroupOpenPhaseParams>`—for every entry in `opening_work` before the first batch is visited. In the ordinary search path, the precommit-opening product is not capped, so this adds `O(opening_work × precommitted_groups)` live state to a planner path that is explicitly intended to remain memory-bounded. Please prepare one work item, reuse it across the inner-basis loop, and then drop it, or process a bounded chunk while preserving the existing ordering.

## Inline comment 3

**File:** `crates/akita-planner/src/planner.rs`  
**Line:** 492  
**Title:** `[P2] Do not retain the obsolete API as a test-only wrapper`

This function reconstructs the old API by calling `PreparedRootProducers::prepare` and immediately forwarding into `root_level_candidates_for_prepared_producers`. It exists only to avoid migrating the remaining test call, while adding another `too_many_arguments` API and a second spelling of the same operation in the production module. Please migrate the test to the prepared-producer API (or place any necessary convenience helper in test support) and keep only the canonical production-layer function.

## Suggested overall review

The preparation reuse is directionally useful, but the current implementation introduces an unreachable duplicate root path, eagerly retains preparation state for the entire opening-work domain, and preserves the superseded API as a test-only production wrapper. These should be addressed before merging so the optimization does not weaken the planner's memory-bounding design or add avoidable maintenance paths.

omibo (approved):

PR #44, opened 2026-09-21
feat: prune unused field tiers at compile time

quangvdao:
## Summary

This PR adds generic compile-time field-tier pruning for Akita's executable PCS stack while preserving the complete protocol policy and default behavior.

- `field-fp32`, `field-fp64`, and `field-fp128` are positive, additive capabilities.
- Default builds still enable all three tiers.
- Unselected dispatch bodies are removed in `akita-types` before downstream typechecking and monomorphization.
- Every canonical ring degree remains available for each enabled tier.
- Unsupported fields fail with `AkitaError::InvalidSetup` at scheme, setup, commitment, prover, verifier, and raw dispatch boundaries.
- Schedule artifacts, proof formats, security tables, and offline planning are unchanged.

### Diff metadata

- Base: `252abb895046cc1d5b9955a26a2ad2318148ac26` (`main`)
- Head: `dc3b87445ef72e37166a47851ec3b224b22a225e`
- Commits: 3
- Files changed: 32
- Diff: 457 insertions, 57 deletions

## Motivation

Akita previously compiled generic dispatch branches for every PCS field tier even when an application used only one. A downstream `cfg(feature = ...)` is insufficient because an exported macro would inspect the consuming crate's features. The selection must happen in the defining crate.

## Design

The canonical dispatch table remains unconditional and is still the single source of truth for field tiers and ring degrees. It routes each tier through an exported helper whose definition is selected inside `akita-types`:

- an enabled helper expands every canonical degree arm for that tier;
- a disabled helper omits the body and returns the shared capability error.

`akita-prover`, `akita-verifier`, `akita-setup`, and `akita-pcs` forward the three capabilities. Their defaults select all tiers. `akita-sis-estimator` and fuzzing opt out of execution defaults so policy-only and narrow workspace graphs are not widened accidentally.

Zero-field builds remain valid for shared types and offline planning. The execution crates also compile in that configuration and reject all PCS fields cleanly; the small explicit type annotations in prover call sites are needed because no enabled macro body remains to infer those result types.

## Change surface

| Area | Result |
| --- | --- |
| Dispatch | Defining-crate helpers prune disabled generic bodies before downstream typechecking |
| Public features | Three additive field capabilities, forwarded through all execution crates |
| Admission | Early field-capability checks at distinct public setup/prove/verify/commit paths |
| Tests | One focused capability test, one downstream compile fixture, one PCS admission test |
| CI and docs | Existing full-support commands select all tiers explicitly; narrow/zero-field behavior is documented and checked |

The downstream fixture deliberately implements only the Fp64 compression degrees 16 and 32. It would fail to compile if disabled Fp32's degree-64 body or Fp128's degree-8 body leaked into downstream typechecking.

## Security and compatibility

- No modulus, CRT representation, transcript, proof/setup encoding, schedule row, or security estimate changes.
- No verifier dependency-boundary change and no new panic path.
- Default/full builds retain Fp32, Fp64, Fp128, and all existing ring degrees.
- Cargo feature unification remains additive: any dependency enabling another tier widens the final binary. Integrators must use `default-features = false` consistently when pruning matters.

## Validation at `dc3b87445`

Focused local checks passed:

- Rust and TOML formatting;
- Fp64-only capability and all-degree dispatch tests;
- additive Fp32+Fp128 capability test;
- zero-field capability/error test;
- downstream Fp64 compile-pruning fixture;
- Fp64 PCS catalog admission test;
- zero-field `akita-pcs` compilation.

An independent read-only review found no production correctness defect or worthwhile remaining deletion after the diff-reduction pass. GitHub CI is rerunning for the current head; this description does not claim it is green yet.

## Scope and remaining work

This PR is based directly on `main`; it does not stack on the planner or backend PRs. It implements the generic Akita mechanism only. Fresh Aerie integration and field-only compilation measurements remain separate downstream work, so this PR makes no performance claim and does not regenerate schedule catalogs.

## Reviewer map

1. `crates/akita-types/src/dispatch/policy.rs` and `dispatch/mod.rs`
2. execution-crate feature forwarding and admission call sites
3. `crates/akita-types/tests/compiled_fields.rs` and the downstream fixture
4. `crates/akita-pcs/tests/compiled_fields.rs`
5. CI command updates and Book recipes


omibo:
## Blockers

### B1. `akita-sis-estimator` no longer compiles — verified locally

`crates/akita-sis-estimator/Cargo.toml` changed to `akita-types = { ..., default-features = false }`. `akita-types`' defaults were the only thing supplying `transcript-blake2b` to that crate's graph, and `akita-sis-estimator` declares no transcript feature of its own.

```
$ cargo check -p akita-sis-estimator
error: enable exactly one transcript backend: transcript-blake2b or transcript-keccak
  --> crates/akita-transcript/src/lib.rs:15:1
```

Reverting only that one line makes it compile again, so the regression is attributable to this PR.

This breaks the offline SIS table regeneration command documented in `AGENTS.md`:

```
cargo run -p akita-sis-estimator --release --features parallel --example infinity_width_table -- ...
```

and roughly fifteen further commands in `scripts/sis_golden/README.md` (`cargo bench -p akita-sis-estimator ...`, `cargo run -p akita-sis-estimator --example infinity_width_table ...`).

No CI job builds `akita-sis-estimator` in isolation — `grep -rn sis-estimator .github/` returns nothing — so workspace feature unification hides this on every green run. Fix: give the crate an explicit transcript feature (or restore the dep), **and** add a `cargo check -p akita-sis-estimator` to CI so the next occurrence is caught.

### B2. The recursion guest still pulls all three tiers — pruning is defeated where it matters most

`profile/akita-recursion/guest/Cargo.toml`:

```toml
akita-types    = { path = "../../../crates/akita-types" }                          # line 47 — default features
akita-verifier = { path = "../../../crates/akita-verifier", default-features = false }  # line 48
```

Line 48 correctly opts out; line 47 does not, so `field-fp32,field-fp64,field-fp128` unify back in and every tier's dispatch body is compiled into the RISC-V guest. `profile/akita-recursion/host` and `.../artifact` have the same unqualified dependency.

### B3. Three of the new tests can pass vacuously

- `crates/akita-pcs/tests/compiled_fields.rs` — the entire file body is `#[cfg(all(feature = "field-fp64", not(feature = "field-fp32"), not(feature = "field-fp128")))]`. Under any other feature set the test binary contains zero tests and exits 0.
- `crates/akita-types/tests/compiled_fields.rs::disabled_fields_report_the_required_feature` — all three assertions are `#[cfg(not(...))]`. Under the default build (all tiers on) the function body is empty and the test passes while asserting nothing.
- `fp64_keeps_every_protocol_ring_degree` and `zero_field_build_has_no_implicit_full_support` are mutually exclusive by construction, so no single run exercises both.

The whole guarantee therefore rests on the seven feature-set invocations in the CI `lint` step being correct and staying correct. If someone reorders a feature list, the tests go quiet rather than red.

Make the vacuity impossible rather than conventional: add a non-`cfg`'d sentinel per file that fails if no expected configuration matched, e.g.

```rust
#[cfg(not(any(feature = "field-fp32", feature = "field-fp64", feature = "field-fp128",
              all(not(feature = "field-fp32"), not(feature = "field-fp64"), not(feature = "field-fp128")))))]
compile_error!("compiled_fields tests must run in a recognised field configuration");
```

or restructure so each file has at least one assertion that holds in every configuration.

---

## Architecture and design

### A1. Six near-identical `#[macro_export]` macros where three `#[cfg]`'d `let` bindings would do

`crates/akita-types/src/dispatch/policy.rs` gains `__dispatch_protocol_policy_fp{32,64,128}` twice each (enabled / `not(enabled)`) — six new macros exported at the crate root, each ~8 lines, differing only in a tier name. The enabled variants are pure forwarding to `__dispatch_ring_dim_arms!`, which is exactly the pass-through indirection `AGENTS.md` ("Single source of truth / no wrapper slop") tells us not to add.

`#[cfg]` on `let` statements is stable, so the same selection collapses into the one existing macro:

```rust
macro_rules! __dispatch_protocol_policy_tiers {
    ($F:ty, $d:expr, |$D:ident| $body:expr;
     fp128: [$($d128:literal),+], fp64: [...], fp32: [...]) => {{
        match $crate::protocol_dispatch_tier::<$F>() {
            $crate::ProtocolRingDispatchTierId::Fp128 => {
                #[cfg(feature = "field-fp128")]
                let out = $crate::__dispatch_ring_dim_arms!($d, $D, $body, { $($d128),+ });
                #[cfg(not(feature = "field-fp128"))]
                let out = { let _ = $d; Err($crate::dispatch::compiled_field_tier_error(
                    $crate::ProtocolRingDispatchTierId::Fp128)) };
                out
            }
            // ... Fp64, Fp32 identically
        }
    }};
}
```

Same behaviour, six fewer exported macros, and the tier table stays readable in one place. Worth confirming the `$body` hygiene holds for the `const $D` binding, but this is the shape to aim for.

### A2. `validate_compiled_field` is scattered across six unrelated call sites, and duplicates a check the dispatch macro already performs

New calls land in `akita-pcs/src/scheme/mod.rs:58`, `akita-prover/src/commitment/api.rs:108`, `akita-prover/src/protocol/core/prove.rs:53`, `akita-setup/src/lib.rs:71`, and `akita-verifier/src/protocol/core/verify.rs:235`. Every one of these paths eventually reaches `dispatch_for_field!`, which returns *the identical* `AkitaError::InvalidSetup` string from `compiled_field_tier_error`. So this is five redundant guards, inserted into five otherwise-unrelated flows — the "scatter feature checks across shared code" pattern.

The list is also ad-hoc rather than principled: `AkitaCommitmentScheme::new` is `pub` and does not validate, so the "admission" story is "some entry points, chosen by hand."

Pick one and state it in a doc comment:

- **(preferred)** rely on the canonical dispatch error and keep at most *one* fail-fast guard where the cost of failing late is real (`akita_setup::setup`, before large allocations). Delete the other four.
- or define "every public constructor and every `pub fn` on `AkitaCommitmentScheme` validates" and enforce it — including `new`.

Either way the current middle ground is the worst of both: redundant, incomplete, and with no rule a future contributor can follow.

### A3. The type-annotation churn is a symptom, not a cost of doing business

Seven call sites acquire explicit types purely because the disabled macro arm carries no type information:

```rust
let witness_source: RecursiveFoldSource<F> = RecursiveFoldSource::witness(...);
let (e_folded, fold_output, extension_opening_reduction): (
    RingVec<F>,
    crate::protocol::fold_grind::TerminalFoldGrindOutput<F>,
    Option<akita_types::ExtensionOpeningReductionProof<E>>,
) = dispatch_for_field!(...)
```

This is inference debt pushed from the definition onto every caller, and it will recur on every future dispatch site and on any tier addition. Two of the annotations (`suffix.rs:299-300`) are not even at dispatch sites — they are collateral from the inference change.

If the annotation is genuinely unavoidable, make it part of the contract instead of an ad-hoc `let`: add a mandatory return-type parameter to `dispatch_for_field!`, e.g. `dispatch_for_field!(slot, F, d, -> RingVec<F>, |D| ...)`. Then the requirement is uniform, checked by the macro, and self-documenting, rather than something a contributor discovers via an inference error under a feature set they do not build. At minimum, import `TerminalFoldGrindOutput` rather than writing a `crate::`-qualified path inside a tuple ascription.

### A4. Crate-wide lint suppression makes a CI gate warning-blind

`akita-types`, `akita-prover`, and `akita-verifier` each gain:

```rust
#![cfg_attr(not(any(feature = "field-fp32", ...)), allow(dead_code, unused_imports, unused_mut, unused_variables))]
```

I verified this is load-bearing — removing it from `akita-types` produces ~20 warnings under `--no-default-features --features transcript-blake2b`. But a crate-root blanket allow is the bluntest possible fix, and CI now runs `cargo check -p akita-pcs --no-default-features --features transcript-blake2b` — exactly the configuration where the allow is active — so that gate proves nothing about hygiene in those crates. `unused_mut` in particular hides real mistakes.

Two better options:

1. Scope the allows to the specific modules/items that go dead, so the rest of each crate keeps its lints. The ~20 warnings are concentrated, not diffuse.
2. Reconsider whether *prover/verifier/setup/pcs* need a zero-field mode at all. Only `akita-planner`, `akita-config`, and `akita-schedules` consume the policy tables without execution, and all three already depend on `akita-types` with `default-features = false`. Keeping zero-field for `akita-types` and emitting `compile_error!` from the execution crates would delete two of the three blanket allows, one CI check, and one test — a strictly smaller design.

### A5. `field-fp32,field-fp64,field-fp128` is hand-copied 15 times

`ci.yml` ×7, `AGENTS.md` ×3, `docs/soundness-audit.md` ×2, `portability.yml` ×1, `docs/ci-test-timing.md` ×1, `book/src/usage/feature-flags.md` ×1. Every future tier means finding all fifteen. Add an aggregate to each execution crate:

```toml
fields-all = ["field-fp32", "field-fp64", "field-fp128"]
default    = ["fields-all", "parallel", "transcript-blake2b"]
```

and replace all fifteen call sites with `fields-all`. This also makes `book/src/usage/feature-flags.md`'s "The normal build uses all five" unnecessary.

---

## Performance and efficiency

### P1. The benefit is unquantified, and I could not reproduce one

The PR states "no binary-size or speedup number is claimed." Given the cost above, that is the central gap. A cold `cargo check -p akita-pcs --all-targets` in two fresh target directories gave **42s user CPU in both** the three-tier and the `field-fp64`-only configuration on this machine — no measurable typecheck win. That is a weak measurement (`cargo check` skips the monomorphization and codegen where the real win should appear), but it means the burden is on the PR to produce the number.

Please include, for `field-fp64`-only vs. default:

- `cargo build --release -p akita-pcs --timings` wall and CPU time,
- stripped `.text` size of the recursion guest binary (after fixing B2),
- guest cycle count from `profile/akita-recursion`, if it moves.

If the win turns out to be small, that reframes the whole cost/benefit discussion — better to know now.

### P2. CI cost is real and sequential

The new `Test field-tier feature selection` step appends **seven** cargo invocations to the `lint` job, each with a distinct feature hash, i.e. seven full rebuilds of the Akita graph on the critical path of the job everyone waits on. One of them (`cargo check --manifest-path crates/akita-types/tests/fixtures/field-fp64/Cargo.toml`) is a nested workspace that fetches `jolt` from git and builds `jolt-field` from scratch.

Move this to its own matrix job so the seven configurations run in parallel and `lint` stays fast.

### P3. `validate_compiled_field` in a per-commitment path

`validate_commitment_geometry` is called per commitment; the check is const-foldable but written as a runtime call. Negligible in practice — noting it only because if you keep A2's guards, `compiled_field_tier_enabled` is already `const fn` and the guard could be a `const { assert }`-style compile-time check at the generic instantiation instead.

---

## Test quality

### T1. `compiled_capabilities_match_positive_features` is tautological

```rust
assert_eq!(compiled_field_tier_enabled(ProtocolRingDispatchTierId::Fp32), cfg!(feature = "field-fp32"));
```

That is the function's body restated as an assertion. It cannot fail unless someone edits both sides. Delete it.

### T2. `assert_fp64_dispatches` duplicates the canonical policy table and the macro's slot dispatch

The expected degree lists (`&[64, 128, 256, 512, 1024, 2048]`, `&[16, 32]`, …) are a second copy of the tables in `policy.rs`, free to drift. And the helper hand-writes five near-identical `match` arms re-implementing slot dispatch, because the macro needs a literal slot token.

Drive the expectations from the canonical source instead — `arms_for_slot` / `slot_dim_supported_for_tier` are already exported for exactly this — so the test checks *agreement with policy* rather than agreement with a transcribed copy of it.

### T3. The downstream fixture is only `cargo check`ed

`crates/akita-types/tests/fixtures/field-fp64/src/main.rs` ends in `assert_eq!(dimension, 32)`, but CI only runs `cargo check --manifest-path ...`, so `main` never executes and the assertion is decoration. The `impl Fp64CompressionDegree for Degree<16>` block is dead — nothing instantiates `Degree<16>`.

Also: `/Cargo.lock` is `.gitignore`d while the manifest has a git dependency on `jolt`. That makes the fixture build unpinned and network-dependent inside CI. Either commit the lock file or make the fixture a normal workspace member with `default-features = false`, which removes the nested workspace, the git fetch, and the separate build entirely.

---

## Smaller cleanups (deslop)

| Location | Issue |
| --- | --- |
| `akita-prover/src/lib.rs:1`, `akita-verifier/src/lib.rs:1` | `// Policy-only builds intentionally omit every executable field dispatch body.` duplicated verbatim; a third variant in `akita-types/src/lib.rs:1`. State it once (e.g. in `feature-flags.md`) and link. |
| `akita-pcs/src/scheme/mod.rs:55-57` | `# Errors` doc gains "Field capability is checked before the artifact bytes are decoded." That is call ordering, not an error contract. Drop the second sentence. |
| `book/src/usage/feature-flags.md` | "The normal build uses all five:" — a count that rots the next time a default feature is added. Say "the defaults". |
| `policy.rs` disabled arms | `let _ = &$d;` — the `&` buys nothing for a `usize` expression; `let _ = $d;` reads plainer. |
| `dispatch/mod.rs` | Three public symbols (`compiled_field_tier_enabled`, `compiled_field_tier_error`, `validate_compiled_field`) for one concept, where `validate_compiled_field` is `enabled` + `error`. `compiled_field_tier_enabled`'s only in-tree consumer is the tautological test in T1; consider making it crate-private. |

---

## What is good

- Putting the selection in `akita-types` is the correct call, and the PR explains *why* consumer-side `cfg` cannot work. That reasoning belongs in a code comment on `__dispatch_protocol_policy_tiers!`, not only in the PR body.
- Keeping the canonical protocol tables complete while pruning only the executable bodies is the right separation: planning and security sizing stay whole.
- Positive/additive feature naming (`field-fp64`, not `no-fp64`) is right for Cargo's unification semantics, and the book section explaining unification is genuinely useful.
- The typed `InvalidSetup` rejection respects the verifier no-panic contract.

---

## Suggested order of work

1. **B1** — fix the `akita-sis-estimator` build and add CI coverage for it.
2. **B2** — pin the three `profile/akita-recursion` manifests; this is also what unblocks P1's numbers.
3. **P1** — publish the compile-time / binary-size measurement. If it is small, revisit A4 option 2 before investing further.
4. **B3, T1–T3** — make the tests non-vacuous and policy-derived.
5. **A1, A5** — collapse the six macros; add `fields-all`.
6. **A2, A3** — settle the admission rule and the dispatch return-type contract.
7. **A4, P2, deslop table.**

quangvdao:
Closed since indeed this does not seem to yield any improvement

